PR #2, opened 2021-03-02
added lend function.

aditya26026:
The lend function has been added. A revert message in withdrawBorrowedAmount had to be changed. Lastly chainLink have changed their directory structure, so had to correct that in PriceOracle.sol


contracts/Pool/Pool.sol
```diff
@@ -164,7 +164,7 @@ contract Pool is ERC20PresetMinterPauserUpgradeable {
         }
         require(
```
prateekreddy, line 165: In the first check of collection and loanStartTime, copy loanStatus into a memory variable and then do the checks.

prateekreddy: loanStatus is being reused again and again, so use that as memory variable.

contracts/Pool/Pool.sol
```diff
@@ -164,7 +164,7 @@ contract Pool is ERC20PresetMinterPauserUpgradeable {
...
             loanStatus == LoanStatus.ACTIVE,
-            "Borrower: Loan is not in ACTIVE state"
+            "Pool::withdrawBorrowedAmount - Loan is not in ACTIVE state"
         );
         uint256 _currentCollateralRatio = getCurrentCollateralRatio();
```
prateekreddy, line 169: Is getCurrentCollateralRatio function complete ?

contracts/Pool/Pool.sol
```diff
@@ -196,8 +196,34 @@ contract Pool is ERC20PresetMinterPauserUpgradeable {
     }
 
 
-    function lend(address _lender, uint256 _amountLent) external {
-        
+    function lend(address _lender, uint256 _amountLent) external payable{
+        require(_amountLent != 0, "Pool::lend - Invalid amount");
```
prateekreddy, line 200: Do we need this check ? What happens if we don't check for this.

contracts/Pool/Pool.sol
```diff
@@ -196,8 +196,34 @@ contract Pool is ERC20PresetMinterPauserUpgradeable {
...
+        require(
+            totalSupply() != borrowAmountRequested,
```
prateekreddy, line 202: Do we need this check ?

contracts/Pool/Pool.sol
```diff
@@ -196,8 +196,34 @@ contract Pool is ERC20PresetMinterPauserUpgradeable {
     }
 
 
-    function lend(address _lender, uint256 _amountLent) external {
-        
+    function lend(address _lender, uint256 _amountLent) external payable{
+        require(_amountLent != 0, "Pool::lend - Invalid amount");
+        require(
```
prateekreddy, line 201: Add a check to see if the pool is in collection period

contracts/Pool/Pool.sol
```diff
@@ -196,8 +196,34 @@ contract Pool is ERC20PresetMinterPauserUpgradeable {
...
+            totalSupply() != borrowAmountRequested,
+            "Pool::lend - Requested amount borrowed"
+        );
+        // Do we need to have a check that restricts the borrower from acting as a lender?
+        uint256 _amount = _amountLent;
+        if(_amountLent.add(totalSupply()) > borrowAmountRequested) {
+            _amount = borrowAmountRequested.sub(totalSupply());
+        }
+        // poolInfo.canBurn[msg.sender] == true;
```
prateekreddy, line 210: remove the commented code

contracts/Pool/Pool.sol
```diff
@@ -196,8 +196,34 @@ contract Pool is ERC20PresetMinterPauserUpgradeable {
...
 
-    function lend(address _lender, uint256 _amountLent) external {
-        
+    function lend(address _lender, uint256 _amountLent) external payable{
+        require(_amountLent != 0, "Pool::lend - Invalid amount");
+        require(
+            totalSupply() != borrowAmountRequested,
+            "Pool::lend - Requested amount borrowed"
+        );
+        // Do we need to have a check that restricts the borrower from acting as a lender?
```
prateekreddy, line 205: We cannot stop this anyways as identities are anon. So check not necessary.

contracts/Pool/Pool.sol
```diff
@@ -196,8 +196,34 @@ contract Pool is ERC20PresetMinterPauserUpgradeable {
...
+        uint256 _amount = _amountLent;
+        if(_amountLent.add(totalSupply()) > borrowAmountRequested) {
+            _amount = borrowAmountRequested.sub(totalSupply());
+        }
+        // poolInfo.canBurn[msg.sender] == true;
+
+        if(borrowAsset == address(0)) {
```
prateekreddy, line 212: Use borrowAsset as a memory variable as it is being read multiple times from storage.

contracts/Pool/Pool.sol
```diff
@@ -196,8 +196,34 @@ contract Pool is ERC20PresetMinterPauserUpgradeable {
...
-        
+    function lend(address _lender, uint256 _amountLent) external payable{
+        require(_amountLent != 0, "Pool::lend - Invalid amount");
+        require(
+            totalSupply() != borrowAmountRequested,
+            "Pool::lend - Requested amount borrowed"
+        );
+        // Do we need to have a check that restricts the borrower from acting as a lender?
+        uint256 _amount = _amountLent;
+        if(_amountLent.add(totalSupply()) > borrowAmountRequested) {
```
prateekreddy, line 207: Use borrowAmountRequested as memory variable.

prateekreddy (changes requested):

contracts/Pool/Pool.sol
aditya26026: Some edges are left to be handled.

contracts/Pool/Pool.sol
aditya26026: Nothing really. I will remove this.

contracts/Pool/Pool.sol
aditya26026: Even this has no real use. Wil remove this.

contracts/Pool/Pool.sol
aditya26026: done

contracts/Pool/Pool.sol
aditya26026: done

contracts/Pool/Pool.sol
aditya26026: done

contracts/Pool/Pool.sol
aditya26026: done

contracts/Pool/Pool.sol
aditya26026: done

contracts/Pool/Pool.sol
prateekreddy: Let's complete  that because we will miss them out if we don't do that now.

contracts/Pool/Pool.sol
```diff
@@ -196,8 +197,30 @@ contract Pool is ERC20PresetMinterPauserUpgradeable {
     }
 
 
-    function lend(address _lender, uint256 _amountLent) external {
-        
+    function lend(address _lender, uint256 _amountLent) external payable{
+        require(loanStatus == LoanStatus.COLLECTION, "Pool::lend - The pool should be in Collection Period.");
+
+        uint256 _amount = _amountLent;
+        uint256 borrowAmountNeeded = borrowAmountRequested;
```
prateekreddy, line 204: Use _ before memory variables, so that  it  is  easy for us to identify them

contracts/Pool/Pool.sol
```diff
@@ -196,8 +197,30 @@ contract Pool is ERC20PresetMinterPauserUpgradeable {
...
+        if(_amountLent.add(totalSupply()) > borrowAmountNeeded) {
+            _amount = borrowAmountNeeded.sub(totalSupply());
+        }
+
+        uint256 borrowToken = borrowAsset;
```
prateekreddy, line 209: Use _ before memory variables, so that it is easy for us to identify them

prateekreddy (changes requested):

contracts/Pool/Pool.sol
aditya26026: will work on this next.

contracts/Pool/Pool.sol
```diff
@@ -139,32 +142,80 @@ contract Pool is ERC20PresetMinterPauserUpgradeable {
     }
 
     // Deposit collateral
-    function deposit(uint256 _amount)
-        external
-        payable
-    {
-        
-    }
+    function deposit(uint256 _amount,bool _isDirect) external payable override {
```
prateekreddy, line 145: Rename to depositCollateral

contracts/Pool/Pool.sol
```diff
@@ -139,32 +142,80 @@ contract Pool is ERC20PresetMinterPauserUpgradeable {
...
+
+        require(_amount != 0, "Pool::deposit - collateral amount");
+        uint256 _sharesReceived;
+        ISavingAccount _savingAccount = ISavingAccount(IPoolFactory(PoolFactory).SavingAccount());
+        address _collateralAsset = collateralAsset;
+        address _investedTo = investedTo;
+        uint256 _liquidityshare = IYield(_investedTo).getTokensForShares(_amount, _collateralAsset);
```
prateekreddy, line 152: Move this to line164 as this is only getting used there.

contracts/Pool/Pool.sol
```diff
@@ -139,32 +142,80 @@ contract Pool is ERC20PresetMinterPauserUpgradeable {
...
+    {
+        require(loanStatus == LoanStatus.ACTIVE, "Pool::addCollateralMarginCall - Loan needs to be in Active stage to deposit"); 
+        require(lenders[_lender].marginCallEndTime >= block.timestamp, "Pool::addCollateralMarginCall - Can't Add after time is completed");
+        require(_amount !=0, "Pool::addCollateralMarginCall - collateral amount");
+
+        uint256 _sharesReceived;
+        ISavingAccount _savingAccount = ISavingAccount(IPoolFactory(PoolFactory).SavingAccount());
+        address _collateralAsset = collateralAsset;
+        address _investedTo = investedTo;
+        uint256 _liquidityshare = IYield(_investedTo).getTokensForShares(_amount, _collateralAsset);
```
prateekreddy, line 182: Move this to line 195

contracts/Pool/Pool.sol
```diff
@@ -139,32 +142,80 @@ contract Pool is ERC20PresetMinterPauserUpgradeable {
...
+
+        if(_isDirect){
+            if(_collateralAsset == address(0)) {
+                require(msg.value == _amount, "Pool::addCollateralMarginCall - value to transfer doesn't match argument");
+                _sharesReceived = _savingAccount.deposit{value:msg.value}(_amount,_collateralAsset,_investedTo, address(this));
+            }
+            else{
+                IERC20(collateralAsset).approve(_investedTo, _amount);
```
prateekreddy, line 190: Approve is not necessary here

contracts/Pool/Pool.sol
```diff
@@ -139,32 +142,80 @@ contract Pool is ERC20PresetMinterPauserUpgradeable {
...
 
 
+    function addCollateralInMarginCall(address _lender,  uint256 _amount,bool _isDirect) external payable override
+    {
+        require(loanStatus == LoanStatus.ACTIVE, "Pool::addCollateralMarginCall - Loan needs to be in Active stage to deposit"); 
+        require(lenders[_lender].marginCallEndTime >= block.timestamp, "Pool::addCollateralMarginCall - Can't Add after time is completed");
+        require(_amount !=0, "Pool::addCollateralMarginCall - collateral amount");
+
+        uint256 _sharesReceived;
+        ISavingAccount _savingAccount = ISavingAccount(IPoolFactory(PoolFactory).SavingAccount());
```
prateekreddy, line 179: If the below logic for deposit is same for depositCollateral and addCollateralInMarginCall, why not seperate it out into a common internal function, something like _depositCollateral.

contracts/Pool/Pool.sol
```diff
@@ -139,32 +142,80 @@ contract Pool is ERC20PresetMinterPauserUpgradeable {
...
         external
-        OnlyBorrower
+        OnlyBorrower override
     {
-        if(loanStatus == LoanStatus.COLLECTION && loanStartTime < block.timestamp) {
+        LoanStatus _poolStatus = loanStatus;
+        if(_poolStatus == LoanStatus.COLLECTION && loanStartTime < block.timestamp) {
             if(totalSupply() < borrowAmountRequested.mul(minborrowAmountFraction).div(100)) {
-                loanStatus = LoanStatus.CANCELLED;
+                _poolStatus = LoanStatus.CANCELLED;
```
prateekreddy, line 211: Should the pool be cancelled in this case ? If loan has started, but the extension period to match collateral is till going on. In that period if user tries to withdraw, then in that case, pool shouldn't be cancelled right, atleast until that collateral match period is complete.

contracts/Pool/Pool.sol
```diff
@@ -139,32 +142,80 @@ contract Pool is ERC20PresetMinterPauserUpgradeable {
...
             }
-            loanStatus = LoanStatus.ACTIVE;
+            _poolStatus = LoanStatus.ACTIVE;
         }
         require(
-            loanStatus == LoanStatus.ACTIVE,
-            "Borrower: Loan is not in ACTIVE state"
+            _poolStatus == LoanStatus.ACTIVE,
+            "Pool::withdrawBorrowedAmount - Loan is not in ACTIVE state"
         );
```
prateekreddy, line 219: I am thinking if we should add a flag to check if amount is borrowed by user. There might be some edgecases like if the loan is repaid and no one actually collected the repayment, in that case also all these conditions will be satisfied and there might be enough amount in the pool for user to withdraw. So this can probably lead to attacks.

prateekreddy: Actually matchCollateralRatioEndTime value can be used instead of a new flag, add a check on this and ensure this is not zero, because this becomes 0 when amount is withdrawn and this is set when pool is created, so we can use this instead

contracts/interfaces/IPool.sol
```diff
@@ -2,163 +2,15 @@
...
-        CLOSED,
-        CANCELLED,
-        DEFAULTED,
-        TERMINATED
-    }
 
-    function lend(address _lender, uint256 _amountLent) external;
 
-    function deposit(address _lender, uint256 _amount) external payable;
+    function deposit(uint256 _amount, bool _isDirect) external payable;
```
prateekreddy, line 7: change to deposit collateral

prateekreddy (changes requested):

contracts/Pool/Pool.sol
aditya26026: This is for the case if say the lenders have not lent out the minimum borrowAmount.

aditya26026: matchCollateral is for the borrowers right.

contracts/Pool/Pool.sol
aditya26026: good point. Will make this change.

contracts/Pool/Pool.sol
aditya26026: I have told @ayush2843 to do this. He will do this.

contracts/Pool/Pool.sol
```diff
@@ -285,8 +287,30 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
     }
 
 
-    function lend(address _lender, uint256 _amountLent) external {
-        
+    function lend(address _lender, uint256 _amountLent) external payable{
+        require(loanStatus == LoanStatus.COLLECTION, "Pool::lend - The pool should be in Collection Period.");
```
prateekreddy, line 291: Also check if the loanStartTime is not already reached. If reached, then revert. Because loanStatus doesn't get updated auto, so we might have to check for this where ever loanStartTime is used.

contracts/Pool/Pool.sol
```diff
@@ -306,32 +330,56 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
         external
         OnlyBorrower
     {   
-        
+        LoanStatus _poolStatus = loanStatus;
+        require(
+            _poolStatus == LoanStatus.COLLECTION || (_poolStatus == LoanStatus.ACTIVE && block.timestamp<matchCollateralRatioEndTime), "Pool::cancelOpenBorrowPool - The pool cannot be cancelled when the status is active."
```
prateekreddy, line 335: block.timestamp<matchCollateralRatioEndTime probably covers all the cases right. This ensures that money is not withdrawn already.

contracts/Pool/Pool.sol
```diff
@@ -306,32 +330,56 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
...
+        );
+        require(matchCollateralRatioEndTime == 0, "Pool::cancelOpenBorrowPool - The borrowedAmount has already been withdrawn.");
```
prateekreddy, line 337: it is != 0, but anyways above require already fixes this. So this require is probably not necessary.

contracts/Pool/Pool.sol
```diff
@@ -306,32 +330,56 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
...
+        loanStatus = LoanStatus.CANCELLED;
+        _pause(); 
+        emit OpenBorrowPoolCancelled();
```
prateekreddy, line 340: When pool is cancelled, the borrower's collateral should be returned after a penalty right.

contracts/Pool/Pool.sol
```diff
@@ -306,32 +330,56 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
...
 
-
-    
     function terminateOpenBorrowPool()
         external
         onlyOwner
     {
-        
+        LoanStatus _poolStatus = loanStatus;
+        require(
```
prateekreddy, line 348: Actually it can be terminated any time.

contracts/Pool/Pool.sol
```diff
@@ -306,32 +330,56 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
...
+            _poolStatus == LoanStatus.ACTIVE || _poolStatus == LoanStatus.COLLECTION,
+            "Pool::terminateOpenBorrowPool - The pool can only be terminated if it is Active or Collection Period."
+        );
+        if (matchCollateralRatioEndTime == 0){
+            emit LoanDefaulted();
```
prateekreddy, line 353: When borrower withdrew or not, the entire collateral should be transferred to the admin right. If user took the lent amount by  then. There is nothing we can do.

contracts/Pool/Pool.sol
```diff
@@ -306,32 +330,56 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
...
+            "Pool::closeLoan - The pool can only be closed if the loan is Active."
+        );
+        uint256 _totalAsset;
+        if (borrowAsset != address(0)) {
+            _totalAsset = IERC20(borrowAsset).balanceOf(address(this));
+        } else {
+            _totalAsset = address(this).balance;
+        }
+        // assuming that the principle is transferred to the pool.
+        require(nextDuePeriod==0 && _totalAsset>totalSupply(), "Pool::closeLoan - The loan has not been fully repayed.");
```
prateekreddy, line 379: _totalAsset can include alot of things right, so this might not be correct.

prateekreddy (changes requested):

contracts/Pool/Pool.sol
aditya26026: right

contracts/Pool/Pool.sol
aditya26026: it seems correct. Will change this.

contracts/Pool/Pool.sol
aditya26026: yeah and removed it.

contracts/Pool/Pool.sol
aditya26026: Not really sure about this. Haven't really discussed this.

contracts/Pool/Pool.sol
aditya26026: Can it be terminated after the pool is Defaulted, Closed or cancelled.

contracts/Pool/Pool.sol
aditya26026: will change this.

contracts/Pool/Pool.sol
aditya26026: Previously I had the thought to put the principle into the pool and the repayments in the repayments contract. So that is why I had used this condition. I am now assuming that after everything is repaid the variable nextDuePeriod would be set as 0.

contracts/Pool/Pool.sol
prateekreddy: Ohh yeah, I mean, it can be terminated when it is active  or collection

prateekreddy (approved):

PR #4, opened 2021-03-02
fixed withdrawAllCollateral

ayush2843:


contracts/Pool/Pool.sol
```diff
@@ -238,11 +238,21 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
         
     }
 
-    function withdrawCollateral()
+    function withdrawAllCollateral()
```
prateekreddy, line 241: Is this the only task that borrower will have to do when the loan is completed

contracts/Pool/Pool.sol
```diff
@@ -238,11 +238,21 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
...
     {
-        
+        LoanStatus _status = loanStatus;
+        require(
+            _status == LoanStatus.CLOSED || _status == LoanStatus.CANCELLED,
+            "Pool::withdrawAllCollateral: Loan is not CLOSED or CANCELLED"
+        );
+
+        uint256 _collateralShares = baseLiquidityShares.add(extraLiquidityShares);
+        uint256 _sharesReceived = ISavingAccount(IPoolFactory(PoolFactory).SavingAccount()).transferFrom(address(this),borrower,_collateralShares,collateralAsset,investedTo);
```
prateekreddy, line 252: This should be transfer right, instead of transferFrom because holder of the tokens is actually transferrinng tokens to someone else.

prateekreddy: borrower can be replace with msg.sender

contracts/Pool/Pool.sol
ayush2843: ya if borrower pays back the principle of the lenders then loan status will be closed and borrower can withdraw all collateral

prateekreddy (changes requested):

contracts/Pool/Pool.sol
ayush2843: fixing this

PR #5, opened 2021-03-02
Sublime v1 loan status updates

aditya26026:
I have added pause() in all status updates. Not really sure if this is needed or if it could indeed cause problems.


contracts/Pool/Pool.sol
```diff
@@ -277,7 +277,13 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
         external
         OnlyBorrower
     {   
-        
+        LoanStatus _poolStatus = loanStatus;
+        require(
+            _poolStatus == LoanStatus.COLLECTION || (_poolStatus == LoanStatus.ACTIVE && block.timestamp<matchCollateralRatioEndTime), "Pool::cancelOpenBorrowPool - The pool cannot be cancelled when the status is active."
```
prateekreddy, line 282: If the borrower matched the collateral, in that case cancel should not be possible right.

contracts/Pool/Pool.sol
```diff
@@ -277,7 +277,13 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
...
+        );
+        loanStatus = LoanStatus.CANCELLED;
+        pause();
```
prateekreddy, line 285: Contract should have pauser role for that. Instead probably call the _pause() method, which is an internal function

contracts/Pool/Pool.sol
```diff
@@ -286,23 +292,38 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
...
         onlyOwner
     {
-        
+        LoanStatus _poolStatus = loanStatus;
+        require(
+            _poolStatus == LoanStatus.ACTIVE || _poolStatus == LoanStatus.COLLECTION,
+            "Pool::terminateOpenBorrowPool - The pool can only be terminated if it is Active or Collection Period."
+        );
+        loanStatus = LoanStatus.TERMINATED;
+        pause();
```
prateekreddy, line 301: Contract should have pauser role for that. Instead probably call the _pause() method, which is an internal function

contracts/Pool/Pool.sol
```diff
@@ -286,23 +292,38 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
...
+        loanStatus = LoanStatus.TERMINATED;
+        pause(); 
+        emit OpenBorrowPoolTerminated();
     }
 
-    // TODO: repay function will invoke this fn
     function closeLoan()
-        internal
-        // onlyOwner // TODO: to be updated  --fixed
+        internal
```
prateekreddy, line 306: Who calls this function and when ?

contracts/Pool/Pool.sol
```diff
@@ -286,23 +292,38 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
...
+            "Pool::closeLoan - The pool can only be closed if the loan is Active."
+        );
+        loanStatus = LoanStatus.CLOSED;
+        pause();
+        emit OpenBorrowPoolClosed();
     }
 
-    // TODO: When repay is missed (interest/principle) call this
     function defaultLoan()
         internal
```
prateekreddy, line 318: Who will call this function and when ?

contracts/Pool/Pool.sol
aditya26026: You are right about this. So I will check if the collateral Ratio crosses the threshold when the function is called

contracts/Pool/Pool.sol
aditya26026: if a loan repayment is missed, then a bot should call a function(yet to be made) which calls this function.

contracts/Pool/Pool.sol
aditya26026: it should be called internally when the borrower repays the full amount. This check should be present in repayAmount.

contracts/Pool/Pool.sol
aditya26026: got it

contracts/Pool/Pool.sol
aditya26026: got it.

contracts/Pool/Pool.sol
prateekreddy: hmm... I am just thinking why do we need a different function for it. Can we club it with repayAmount only, or is it in a different contract.

contracts/Pool/Pool.sol
prateekreddy: Hmm... but it doesn't have any checks to ensure that the loan payment was really missed. Also this is an internal function

contracts/Pool/Pool.sol
prateekreddy: If it  is in different contract, then this function can't be external, but again  it cannot be public as well because not everyone should be  able to  call this function.

prateekreddy:
> I have added pause() in all status updates. Not really sure if this is needed or if it could indeed cause problems.

I think this is fine, when the loan is not active (for whatever reason like closed, terminated, cancelled), then there is no point in transferring lending positions to someone else. So it makes sense to pause all lending position transfers.

contracts/Pool/Pool.sol
aditya26026: We can. So i will remove the function itself.

contracts/Pool/Pool.sol
aditya26026: Should _pause() be called for in case if a loan is terminated, defaulted or cancelled.

contracts/Pool/Pool.sol
aditya26026: So, there has to be a function which checks if a repayment is missed and then calls this defaultLoan function.

prateekreddy: sure, let's add this to the liquidation function I guess.

contracts/Pool/Pool.sol
```diff
@@ -277,32 +277,49 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
...
         onlyOwner
     {
-        
+        LoanStatus _poolStatus = loanStatus;
+        require(
+            _poolStatus == LoanStatus.ACTIVE || _poolStatus == LoanStatus.COLLECTION,
+            "Pool::terminateOpenBorrowPool - The pool can only be terminated if it is Active or Collection Period."
+        );
+        if (matchCollateralRatioEndTime == 0){
+            emit LoanDefaulted();
```
prateekreddy, line 300: If user has already withdrawn the collateral by the time loan is terminated, what is the course of action ?

contracts/Pool/Pool.sol
```diff
@@ -277,32 +277,49 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
...
-        // onlyOwner // TODO: to be updated
+        external   
+        OnlyBorrower
     {
-        
+        require(
+            loanStatus == LoanStatus.ACTIVE,
+            "Pool::closeLoan - The pool can only be closed if the loan is Active."
+        );
+        // TODO: check if the loan has beencompletely repayed.
```
prateekreddy, line 319: Let's merge once this is complete

prateekreddy (changes requested):

contracts/Pool/Pool.sol
aditya26026: how could the user withdraw collateral if the loan is in active status

contracts/Pool/Pool.sol
aditya26026: The borrower would call the function when the loan is repayed and he wants to close the loan.

contracts/Pool/Pool.sol
```diff
@@ -316,7 +316,14 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
...
         );
-        // TODO: check if the loan has beencompletely repayed.
+        uint256 _totalAsset;
+        if (borrowAsset != address(0)) {
+            _totalAsset = IERC20(borrowAsset).balanceOf(address(this));
+        } else {
+            _totalAsset = address(this).balance;
+        }
+        // assuming that the principle is transferred to the pool.
+        require(nextDuePeriod==0 && _totalAsset>totalSupply(), "Pool::closeLoan - The loan has not been fully repayed.");
```
aditya26026, line 326: nextDuePeriod would start out with 1 and we would set it to 0 if all interest repayments are done. I have assumed here that the principle would be transferred to the pool contract and not the repayment Contract.

prateekreddy (approved):

PR #6, opened 2021-03-02
Sublime v1 utility functions

aditya26026:
Functions like amountPerPeriod, interestPerSecond, amountLenderPerPeriod and calculateCurrentPeriod

### DIFF
--- a/contracts/Pool/Pool.sol
+++ b/contracts/Pool/Pool.sol
@@ -391,7 +391,7 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
         return _interest;
     }
     
-    function amountLenderPerPeriod(address _lender)
+    function interestPerPeriod(address _lender)
         public
         view
         returns (uint256)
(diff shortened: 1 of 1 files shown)


contracts/Pool/Pool.sol
```diff
@@ -382,16 +382,31 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
...
+    function interestPerSecond(uint256 _principle)
+        public
+        view
+        returns (uint256)
+    {
+        uint256 _interest = ((_principle).mul(borrowRate)).div(365 days);
+        return _interest;
+    }
+    
+    function amountLenderPerPeriod(address _lender)
```
prateekreddy, line 394: Can you rename the function ? Intent is not very clear

contracts/Pool/Pool.sol
```diff
@@ -382,16 +382,31 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
...
+        returns (uint256)
+    {
+        return (interestPerSecond(balanceOf(_lender)).mul(repaymentInterval));
     }
 
-    function amountLenderPerPeriod(address lender) public view returns(uint256){
-        
+    function calculateCurrentPeriod() public view returns (uint256) {
+        uint256 _currentPeriod =
+            (block.timestamp.sub(loanStartTime)).div(repaymentInterval);
```
prateekreddy, line 404: Give the error message in sub as well, basically "Loan hasn't started yet" or something on those lines

aditya26026: will do.

contracts/Pool/Pool.sol
aditya26026: will do.

aditya26026: I have changed both amountPerPeriod and amountLenderPerPeriod to an overloaded interestPerPeriod

PR #7, opened 2021-03-02
Sublime v1 collateral ratio

aditya26026:
Added an extra function so that the stack does not get too deep.

### DIFF
--- a/contracts/Pool/Pool.sol
+++ b/contracts/Pool/Pool.sol
@@ -369,7 +369,7 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
         
     }
 
-    function interestTillNow(uint256 balance, uint256 _interestPerPeriod) public returns(uint256){
+    function interestTillNow(uint256 _balance, uint256 _interestPerPeriod) public returns(uint256){
         uint256 _repaymentLength = repaymentInterval;
         uint256 _loanStartedAt = loanStartTime;
         uint256 _totalSupply = totalSupply();
@@ -385,7 +385,7 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
                 )
             );
         uint256 _extraInterest =
-            interestPerSecond(balance).mul(
+            interestPerSecond(_balance).mul(
                 ((calculateCurrentPeriod().add(1)).mul(_repaymentLength))
                     .add(_loanStartedAt)
                     .sub(block.timestamp)
@@ -402,8 +402,8 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
         }
     }
 
-    function calculateCollateralRatio(uint256 _interestPerPeriod, uint256 balance, uint256 _liquidityShares) public returns(uint256){
-        uint256 _interest = interestTillNow(balance, _interestPerPeriod);
+    function calculateCollateralRatio(uint256 _interestPerPeriod, uint256 _balance, uint256 _liquidityShares) public returns(uint256){
+        uint256 _interest = interestTillNow(_balance, _interestPerPeriod);
         address _collateralAsset = collateralAsset;
         uint256 _ratioOfPrices =
             IPriceOracle(IPoolFactory(PoolFactory).priceOracle())
@@ -414,7 +414,7 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
                 _collateralAsset
             );
         uint256 _ratio = (_currentCollateralTokens.mul(_ratioOfPrices).div(100000000)).div(
-            balance.add(_interest)
+            _balance.add(_interest)
         );
         return(_ratio);
     }
--- a/contracts/Pool/Pool.sol
+++ b/contracts/Pool/Pool.sol
@@ -51,7 +51,7 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
     uint256 public repaymentInterval;
     address public collateralAsset;
     
-    uint256 PeriodWhenExtensionIsRequested;
+    uint256 public periodWhenExtensionIsRequested;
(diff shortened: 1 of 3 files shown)


prateekreddy:
Copy the storage variable into memory, if you are going to use it multiple times.

prateekreddy (changes requested):

contracts/Pool/Pool.sol
```diff
@@ -354,18 +354,101 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
...
+                collateralAsset
+            );
+
+        uint256 _ratioOfPrices =
+            IPriceOracle(IPoolFactory(PoolFactory).priceOracle())
+                .getLatestPrice(collateralAsset, borrowAsset);
+
+        (uint256 _interest, uint256 _gracePeriodsTaken) =
+            (
+                IRepayment(Repayment).calculateRepayAmount(
```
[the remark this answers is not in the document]
aditya26026: This has to be done

contracts/Pool/Pool.sol
```diff
@@ -354,18 +354,101 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
...
+            gracePenaltyRate.mul(amountPerPeriod()).div(100).mul(
+                _gracePeriodsTaken
+            )
+        );
+        if (_interest < _extraInterest) {
+            _interest = 0;
+        } else {
+            _interest = _interest.sub(_extraInterest);
+        }
+        _ratio = (_currentCollateralTokens.mul(_ratioOfPrices)).div(
```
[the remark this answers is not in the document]
aditya26026: ChainLink multiplies the ratios by a factor of 10^8. So I have divided the numbers by the same

contracts/Pool/Pool.sol
```diff
@@ -62,7 +62,7 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
     uint256 public extensionVoteEndTime;
     uint256 public noOfGracePeriodsTaken;
     uint256 nextDuePeriod;
-
+    uint256 gracePenaltyRate;
```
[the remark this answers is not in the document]
aditya26026: done

contracts/Pool/Pool.sol
```diff
@@ -354,18 +354,101 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
     event Liquidated(address liquidator, address lender);
 
     // TODO
```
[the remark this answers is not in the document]
aditya26026: done

contracts/Pool/Pool.sol
```diff
@@ -61,8 +61,8 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
     mapping(address => LendingDetails) public lenders;
     uint256 public extensionVoteEndTime;
     uint256 public noOfGracePeriodsTaken;
-    uint256 nextDuePeriod;
-
+    uint256 public nextDuePeriod;
+    uint256 public gracePeriodPenaltyFraction;
```
prateekreddy, line 65: This will be part of poolFactory

contracts/Pool/Pool.sol
```diff
@@ -353,19 +353,82 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
...
+                ((calculateCurrentPeriod().add(1)).mul(_repaymentLength))
+                    .add(_loanStartedAt)
+                    .sub(block.timestamp)
+            );
+        _interest = _interest.sub(
+            gracePeriodPenaltyFraction.mul(_interestPerPeriod).div(100).mul(
+                _gracePeriodsTaken
+            )
+        );
+        if (_interest < _extraInterest) {
```
prateekreddy, line 384: Is this possible ? Does interest already include the extraInterest as well right

contracts/Pool/Pool.sol
```diff
@@ -353,19 +353,82 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
...
-    // TODO
     function getCurrentCollateralRatio(address _lender)
         public
-        returns (uint256 ratio) {
-
+        returns (uint256 _ratio)
+    {
+        uint256 _balanceOfLender = balanceOf(_lender);
+        uint256 _totalSupply = totalSupply();
+        address _collateralToken = collateralAsset;
```
prateekreddy, line 418: If both use similar code, then can you seperate the common code into a different internal function

prateekreddy: I mean getCurrentCollateralRatio

prateekreddy (changes requested):

contracts/Pool/Pool.sol
aditya26026: will do

contracts/Pool/Pool.sol
aditya26026: Say he has to pay 100 dai in 8 days. The borrower has already paid 60 dai for an interval. The _interest left to be paid would be 40 dai for that interval. On the 4th day, the _extraInterest would be 50 dai. So in that case the borrower does not owe money to the pool.

contracts/Pool/Pool.sol
aditya26026: The problem with having this as a global variable is that changing this during the execution of a pool would mess up all the calculations for liquidation and repayments.

prateekreddy: ok sure, will intialize the pool with the value, in that case

contracts/Pool/Pool.sol
prateekreddy: cool, makes sense.

prateekreddy (approved):

PR #8, opened 2021-03-03
Voting

ayush2843:


contracts/Pool/Pool.sol
```diff
@@ -317,21 +317,27 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
     }
 
 
-    function resultOfVoting() external {
-        
+    function resultOfVoting() external isPoolActive{
+
+        (PeriodWhenExtensionIsPassed,nextDuePeriod) = IRepayment(Repayment).resultOfVoting(totalExtensionSupport, extensionVoteEndTime, totalSupply(),nextDuePeriod,repaymentInterval,loanStartTime,PeriodWhenExtensionIsPassed);
```
prateekreddy, line 322: This function looks incomplete

contracts/Pool/Pool.sol
```diff
@@ -317,21 +317,27 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
...
     }
 
     function requestExtension() external OnlyBorrower isPoolActive
     {
+        require(PeriodWhenExtensionIsPassed > noOfRepaymentIntervals,"Pool::requestExtension: you have already been given an extension,No more extension");
+        extensionVoteEndTime = IRepayment(Repayment).requestExtension(extensionVoteEndTime);
```
prateekreddy, line 328: Why is extensionVoteEndTime passed as argument to requestExtension ?

contracts/Pool/Pool.sol
```diff
@@ -317,21 +317,27 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
...
+        totalExtensionSupport = 0;
```
prateekreddy, line 329: Wouldn't this already be 0 ?

contracts/Pool/Pool.sol
```diff
@@ -317,21 +317,27 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
...
         
     }
 
-
     function voteOnExtension() external isPoolActive 
     {
-        
+        require(balanceOf(msg.sender)!=0 , "Pool::voteOnExtension - insufficient balance for vote");
+        (lenders[msg.sender].lastVoteTime,totalExtensionSupport) = IRepayment(Repayment).voteOnExtension(msg.sender,lenders[msg.sender].lastVoteTime,extensionVoteEndTime, balanceOf(msg.sender),totalExtensionSupport);
+
```
prateekreddy, line 337: May be emit an event or something like that

prateekreddy: or is it being emitted in repayments contract

contracts/Repayments/Repayments.sol
```diff
@@ -34,32 +39,41 @@ contract Repayments is RepaymentStorage {
...
+            (block.timestamp.sub(_loanStartTime)).div(_repaymentInterval);
+        return _currentPeriod;
     }
 
     function interestPerSecond(uint256 _principle, uint256 _borrowRate)
         public
         view
         returns (uint256)
     {
+        uint256 _interest = ((_principle).mul(_borrowRate)).div(365 days);
```
prateekreddy, line 61: The decimals might be an issue here ? Accuracy will be less, so multiply with some constant and then divide later.

contracts/Repayments/Repayments.sol
```diff
@@ -34,32 +39,41 @@ contract Repayments is RepaymentStorage {
...
-    function amountPerPeriod(
+    function amountPerPeriodBorrower(
         uint256 _activeBorrowAmount,
         uint256 _repaymentInterval,
         uint256 _borrowRate
     ) public view returns (uint256) {
-        
+
+        uint256 _amountPerPeriod =
+            interestPerSecond(_activeBorrowAmount, _borrowRate).mul(
```
prateekreddy, line 73: divide by the multiplication factor of interestPerSecond.

contracts/Pool/Pool.sol
```diff
@@ -317,21 +317,27 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
...
-    function resultOfVoting() external {
-        
+    function resultOfVoting() external isPoolActive{
+
+        (PeriodWhenExtensionIsPassed,nextDuePeriod) = IRepayment(Repayment).resultOfVoting(totalExtensionSupport, extensionVoteEndTime, totalSupply(),nextDuePeriod,repaymentInterval,loanStartTime,PeriodWhenExtensionIsPassed);
     }
 
     function requestExtension() external OnlyBorrower isPoolActive
     {
+        require(PeriodWhenExtensionIsPassed > noOfRepaymentIntervals,"Pool::requestExtension: you have already been given an extension,No more extension");
```
prateekreddy, line 327: Check pool status as well

contracts/Repayments/Repayments.sol
```diff
@@ -90,29 +104,66 @@ contract Repayments is RepaymentStorage {
     //     return(intervalLeft.mul(amountPerPeriod()));
     // }
 
-    function requestExtension(uint256 extensionVoteEndTime)
-        external isPoolInitialized
+
+
+    function requestExtension(uint256 _extensionVoteEndTime)
+        external override isPoolInitialized
```
prateekreddy, line 110: Can this be called by anyone or only pool contract

contracts/Repayments/Repayments.sol
```diff
@@ -90,29 +104,66 @@ contract Repayments is RepaymentStorage {
...
         returns (uint256)
     {
-        
+        require(
+            block.timestamp > _extensionVoteEndTime,
```
prateekreddy, line 114: this is being checked in pool contract right.

contracts/Repayments/Repayments.sol
```diff
@@ -90,29 +104,66 @@ contract Repayments is RepaymentStorage {
...
+            "Repayments::requestExtension - Extension requested already"
+        );
+        _extensionVoteEndTime = (block.timestamp).add(repaymentDetails[msg.sender].votingExtensionlength);
```
prateekreddy, line 117: Isn't votingExtensionlength a global variable ?

contracts/Pool/Pool.sol
prateekreddy: will checking if extensionVoteEndTime is non zero not solve no more extension stuff ?

contracts/Repayments/Repayments.sol
prateekreddy: Only pool contract can initialize repayments ? is that why this will work ?

contracts/Repayments/Repayments.sol
```diff
@@ -90,29 +104,66 @@ contract Repayments is RepaymentStorage {
...
-        uint256 balance,
-        uint256 totalExtensionSupport
-    ) external isPoolInitialized returns (uint256, uint256) {
-        
+        address _lender,
+        uint256 _lastVoteTime,
+        uint256 _extensionVoteEndTime,
+        uint256 _balance,
+        uint256 _totalExtensionSupport
+    ) external override isPoolInitialized returns (uint256, uint256) {
```
prateekreddy, line 128: isPoolInitialized sufficient ? Only pool contract can initialize repayments ? is that why this will work ?

contracts/Repayments/Repayments.sol
```diff
@@ -90,29 +104,66 @@ contract Repayments is RepaymentStorage {
...
+        require(
+            block.timestamp < _extensionVoteEndTime,
+            "Repayments::voteOnExtension - Voting is over"
+        );
+        require(
+            _lastVoteTime < _extensionVoteEndTime.sub(repaymentDetails[msg.sender].votingExtensionlength),
```
prateekreddy, line 134: isn't votingExtensionLength a global variable ?

prateekreddy: rename votingExtensionLength to extensionVotePeriod

prateekreddy: Also what does this check do ? Why will lastVote time be before the extensionVoting started ?

contracts/Repayments/Repayments.sol
```diff
@@ -90,29 +104,66 @@ contract Repayments is RepaymentStorage {
...
+            "Repayments::voteOnExtension - you have already voted"
+        );
+        _lastVoteTime = block.timestamp;
```
prateekreddy, line 137: Why is _lastVoteTime being used ?

contracts/Repayments/Repayments.sol
```diff
@@ -90,29 +104,66 @@ contract Repayments is RepaymentStorage {
...
 
     function voteOnExtension(
-        address lender,
-        uint256 lastVoteTime,
-        uint256 extensionVoteEndTime,
-        uint256 balance,
-        uint256 totalExtensionSupport
-    ) external isPoolInitialized returns (uint256, uint256) {
-        
+        address _lender,
```
prateekreddy, line 123: Is this function necessary here, most of the logic is based on variables in the other contract only.

contracts/Repayments/Repayments.sol
```diff
@@ -90,29 +104,66 @@ contract Repayments is RepaymentStorage {
...
+        uint256 _totalSupply,
+        uint256 _nextDuePeriod,
+        uint256 _repaymentInterval,
+        uint256 _loanStartTime,
+        uint256 _PeriodWhenExtensionIsPassed
+    ) external override isPoolInitialized returns (uint256,uint256) {
+
+        require(block.timestamp > _extensionVoteEndTime, "Repayments::resultOfVoting - Voting is not over");
+
+        // Assuming votingPassRatio a uint in range of 1-100 (can be changed to 10^18 or some large value)
```
prateekreddy, line 156: Make votingPassRatio to use 10^8 accuracy.

contracts/Repayments/Repayments.sol
```diff
@@ -90,29 +104,66 @@ contract Repayments is RepaymentStorage {
...
-        uint256 nextDuePeriod
-    ) external isPoolInitialized returns (uint256) {
+        uint256 _totalExtensionSupport,
+        uint256 _extensionVoteEndTime,
+        uint256 _totalSupply,
+        uint256 _nextDuePeriod,
+        uint256 _repaymentInterval,
+        uint256 _loanStartTime,
+        uint256 _PeriodWhenExtensionIsPassed
+    ) external override isPoolInitialized returns (uint256,uint256) {
```
prateekreddy, line 152: Why is this function in repayments when none of the variables in repayments contract are being used here.

contracts/Repayments/Repayments.sol
```diff
@@ -90,29 +104,66 @@ contract Repayments is RepaymentStorage {
...
+
+        require(block.timestamp > _extensionVoteEndTime, "Repayments::resultOfVoting - Voting is not over");
+
+        // Assuming votingPassRatio a uint in range of 1-100 (can be changed to 10^18 or some large value)
+        if (((_totalExtensionSupport).mul(repaymentDetails[msg.sender].votingPassRatio)).div(100) >= _totalSupply) {
```
prateekreddy, line 157: Isn't voting pass ratio a global variable ?

prateekreddy: Shouldn't the logic here be totalSupply*(votingPassRatio) < totalExtensionSupport

contracts/Repayments/Repayments.sol
prateekreddy: I guess only calculateCurrentPeriod is being used here for which you can do a cross contract call.

prateekreddy (changes requested):

contracts/Pool/Pool.sol
```diff
@@ -51,7 +51,7 @@ contract Pool is ERC20PresetMinterPauserUpgradeable,IPool {
     uint256 public repaymentInterval;
     address public collateralAsset;
     
-    uint256 PeriodWhenExtensionIsRequested;
+    uint256 public PeriodWhenExtensionIsPassed;   // will be set to noOfRepaymentIntervals+1
```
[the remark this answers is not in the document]
ayush2843: done

contracts/Pool/Pool.sol
ayush2843: calling result of voting updates periodWhenExtensionIsPassed and nextDuePeriod if votes are passed and also emits suitable event. I haven't return anything like true or false for vote passing as it wasn't required in the contracts and in the frontend we can use events.

contracts/Pool/Pool.sol
ayush2843: so that borrower doesn't ask for another extension whenever a extension is already running

contracts/Pool/Pool.sol
ayush2843: initially it will be zero but borrower can request Extension multiple times(in case extensions keep on failing) so the totalExtensionSupport has to set to 0 in a new extension request

contracts/Pool/Pool.sol
ayush2843: ya all the events are emitted in repayments

contracts/Pool/Pool.sol
ayush2843: > Check pool status as well

isPoolActive Modifiers checks it

ayush2843: > will checking if extensionVoteEndTime is non zero not solve no more extension stuff ?

Actually borrower can ask for multiple extensions and only one but if extension is passed then it cannot ask for more extension, request can be multiple but only one is passed, that's why extensionVoteEndTime alone will not work for this

contracts/Repayments/Repayments.sol
ayush2843: This check is different from the check in pool, it checks for no more extensions when an extension is already requested. pool check is for the case when an extension is already passed so no more extensions has to be given

contracts/Repayments/Repayments.sol
ayush2843: ya it is global, needs to update this

contracts/Repayments/Repayments.sol
ayush2843: ya it changes and access variables of the msg.sender, and also returns updated value, if pool doesn't call this then no variables will be updated in the pool

contracts/Repayments/Repayments.sol
ayush2843: > Isn't voting pass ratio a global variable ?

ya will update

ayush2843: > Shouldn't the logic here be totalSupply*(votingPassRatio) < totalExtensionSupport

ya true, will update

contracts/Repayments/Repayments.sol
ayush2843: ya will move it to the pool

contracts/Repayments/Repayments.sol
ayush2843: sure

PR #10, opened 2021-03-04
Savings account

ameesha1205:
Added functions relevant for savings account
Also updated yields and interfaces as per requirement


contracts/yield/StrategyRegistry.sol
```diff
@@ -2,26 +2,106 @@
...
+        require(
+            _maxStrategies != 0,
+            "StrategyRegistry::updateMaxStrategies should be more than zero"
+        );
+        maxStrategies = _maxStrategies;
     }
 
-    function addStrategy(address _strategy) external onlyOwner {
-        registry[_strategy] = true;
+    function registry(address strategy) external view override returns (bool) {
```
prateekreddy, line 41: Probably storing a mapping is better, because these strategies don't change often, but this function is used everytime a new pool is created, so it might actually overall save more gas if these are also stored in mapping.

prateekreddy (changes requested):

contracts/SavingsAccount/SavingsAccount.sol
```diff
@@ -0,0 +1,339 @@
...
+
+        uint256 tokensReceived = amount;
+        if (currentStrategy != address(0)) {
+            tokensReceived = IYield(currentStrategy).unlockTokens(
+                asset,
+                amount
+            );
+        }
+
+        uint256 sharesReceived;
```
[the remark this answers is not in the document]
ameesha1205: Yes, updated

contracts/yield/StrategyRegistry.sol
ameesha1205: Okay, but in withdrawAll(), it will be tough to identify and withdraw from strategies. Also if we are planning to get total investments in a particular asset, this can help. But if we are going to have them in the global registry we are planning, I will update this.

PR #25, opened 2021-03-06
Voting updated

ayush2843:


contracts/Pool/Pool.sol
```diff
@@ -554,6 +557,55 @@ contract Pool is ERC20PresetMinterPauserUpgradeable, IPool {
...
+    function requestExtension() external isPoolActive OnlyBorrower
+    {
+        uint256 _extensionVoteEndTime = extensionVoteEndTime;
+        require(
+            block.timestamp > _extensionVoteEndTime,
+            "Pool::requestExtension - Extension requested already"
+        );
+
+        // This check is required so that borrower doesn't ask for more extension if previously an extension is already granted
+        require(periodWhenExtensionIsPassed > noOfRepaymentIntervals,"Pool::requestExtension: you have already been given an extension,No more extension");
```
prateekreddy, line 569: If this is non zero that should be fine ? periodWhenExtensionIsPassed

contracts/Pool/Pool.sol
```diff
@@ -554,6 +557,55 @@ contract Pool is ERC20PresetMinterPauserUpgradeable, IPool {
...
+            "Pool::voteOnExtension - you have already voted"
+        );
+        _lastVoteTime = block.timestamp;
+        totalExtensionSupport = totalExtensionSupport.add(balanceOf(msg.sender));
+        uint256 _votingPassRatio = IPoolFactory(PoolFactory).votingPassRatio();
+        lenders[msg.sender].lastVoteTime = _lastVoteTime;
+        emit lenderVoted(msg.sender,totalExtensionSupport,_lastVoteTime);
+        
+        if (((totalExtensionSupport)) >= (totalSupply().mul(_votingPassRatio)).div(100000000)) {
+            periodWhenExtensionIsPassed = calculateCurrentPeriod();
```
prateekreddy, line 603: Is gracePeriod considered as part of the currentPeriod only, or does it overflow into the next period

prateekreddy: If extension is approved in grace Period, the currentPeriod should be incremented right ? Also how is grace Penality being detected ?

contracts/Pool/Pool.sol
ayush2843: Zero will not work as voting can be passed in period zero also.

prateekreddy (approved):

PR #26, opened 2021-03-07
Lending position transfers

ameesha1205:
Added transfer and transferFrom function in pool

### DIFF
--- a/contracts/Pool/Pool.sol
+++ b/contracts/Pool/Pool.sol
@@ -51,7 +51,7 @@ contract Pool is ERC20PresetMinterPauserUpgradeable, IPool {
     address public collateralAsset;
 
     uint256 public liquidatorRewardFraction;
-    
+
     uint256 public periodWhenExtensionIsPassed;
     uint256 public baseLiquidityShares;
     uint256 public extraLiquidityShares;
@@ -86,7 +86,7 @@ contract Pool is ERC20PresetMinterPauserUpgradeable, IPool {
     //     uint256 sharesReceived
     // );
     event CollateralWithdrawn(address user, uint256 amount);
-    event liquiditySupplied(uint256 amountSupplied, address lenderAddress);
+    event liquiditySupplied(uint256 amountSuppliised, address lenderAddress);
     event AmountBorrowed(address borrower, uint256 amount);
     event Liquiditywithdrawn(uint256 amount, address lenderAddress);
     event CollateralCalled(address lenderAddress);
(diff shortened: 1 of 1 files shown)


contracts/Pool/Pool.sol
```diff
@@ -86,7 +86,7 @@ contract Pool is ERC20PresetMinterPauserUpgradeable, IPool {
     //     uint256 sharesReceived
     // );
     event CollateralWithdrawn(address user, uint256 amount);
-    event liquiditySupplied(uint256 amountSupplied, address lenderAddress);
+    event liquiditySupplied(uint256 amountSuppliised, address lenderAddress);
```
prateekreddy, line 89: Typo

contracts/Pool/Pool.sol
```diff
@@ -416,30 +426,76 @@ contract Pool is ERC20PresetMinterPauserUpgradeable, IPool {
...
     }
 
-    function _beforeTransfer(address _user) internal {}
+    function _beforeTransfer(
+        address _from,
+        address _to,
+        uint256 _amount
+    ) internal {
+        require(
+            lenders[_from].marginCallEndTime != 0,
```
prateekreddy, line 437: This also applies for _to also, because if a margin call is made by the _to address, it would be hard to manage margin call for the _to address after transfer right.

contracts/Pool/Pool.sol
```diff
@@ -414,30 +428,76 @@ contract Pool is ERC20PresetMinterPauserUpgradeable, IPool {
...
+            .add(toTransfer);
+    }
 
     function transfer(address _recipient, uint256 _amount)
         public
         override
         returns (bool)
-    {}
+    {
+        require(!paused(), "Pool::transfer - token transfer while paused");
```
prateekreddy, line 468: This probably is not needed because _transfer already has this internally.

contracts/Pool/Pool.sol
```diff
@@ -414,30 +428,76 @@ contract Pool is ERC20PresetMinterPauserUpgradeable, IPool {
...
+        _transfer(_msgSender(), _recipient, _amount);
+        return true;
+    }
 
     function transferFrom(
         address _sender,
         address _recipient,
         uint256 _amount
-    ) public virtual override returns (bool) {}
+    ) public virtual override returns (bool) {
```
prateekreddy, line 478: Why is this virtual ?

contracts/Pool/Pool.sol
```diff
@@ -414,30 +428,76 @@ contract Pool is ERC20PresetMinterPauserUpgradeable, IPool {
...
+        require(!paused(), "Pool::transferFrom - token transfer while paused");
```
prateekreddy, line 479: This is part of _transfer, so probably not needed.

contracts/Pool/Pool.sol
ameesha1205: Okay adding this

contracts/Pool/Pool.sol
ameesha1205: Right, removed

contracts/Pool/Pool.sol
ameesha1205: Right, removed

contracts/Pool/Pool.sol
ameesha1205: My bad, removed

PR #27, opened 2021-03-07
Deposit to function

ameesha1205:
Added function to deposit to another user's savings account


contracts/SavingsAccount/SavingsAccount.sol
```diff
@@ -79,9 +101,25 @@ contract SavingsAccount is ISavingsAccount, Initializable, OwnableUpgradeable {
         address strategy,
         address user
     ) external payable override returns (uint256 sharesReceived) {
+        sharesReceived = _deposit(amount, asset, strategy, user);
```
prateekreddy, line 104: Deposit function is also attackable, if msg.sender is someone else. Same attack vector as depositTo.

prateekreddy (changes requested):

PR #28, opened 2021-03-07
Adding withdraw repayments functionality

ritik99:
Functions to be reviewed:
1. calculateRepaymentWithdrawable() and withdrawRepayment in Pool.sol
2. getTotalRepaidAmount() in Repayments.sol

There are also two definitions of interestPerPeriod() in Pool.sol. Don't know which one is correct so kept both for now


contracts/Pool/Pool.sol
```diff
@@ -832,6 +833,67 @@ contract Pool is ERC20PresetMinterPauserUpgradeable, IPool {
...
+
+    }
+
+    
+    // Withdraw Repayment, Also all the extra state variables are added here only for the review
+    
+    function withdrawRepayment(bool _withdrawToSavingsAccount) external isLender(msg.sender) {
+        LoanStatus _loanStatus = loanStatus;
+
+        require(_loanStatus != LoanStatus.CANCELLED,
```
prateekreddy, line 859: This can probably be skipped as amountToWithdraw will anyways be 0

PR #33, opened 2021-03-10
loan repay Extension structure added  to extension contract

prateekreddy:


contracts/interfaces/IPool.sol
```diff
@@ -21,4 +21,5 @@ interface IPool {
     ) external ;
 
     function setPoolToken(address _poolToken) external;
+    function borrower() external returns(address)
```
ritik99, line 24: semi-colon

contracts/Pool/Extension.sol
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    function voteOnExtension(address _pool) external isPoolActive {
+        uint256 _extensionVoteEndTime = poolInfo[_pool].extensionVoteEndTime;
+        require(
+            block.timestamp < _extensionVoteEndTime,
+            "Pool::voteOnExtension - Voting is over"
+        );
+        // TODO: This belongs in pool
+        uint256 _balance = poolToken.balanceOf(msg.sender);
```
ritik99, line 75: poolToken will require an interface right?

contracts/Pool/Extension.sol
```diff
@@ -0,0 +1,119 @@
...
+        // TODO: Create a fn in Pool for this and call that fn instead
+        uint256 _nextDueTime = pool.getNextDuePeriod();
+        // uint256 _nextDueTime =
+        //     (poolVars.nextDuePeriod.mul(poolInfo.repaymentInterval)).add(poolConstants.loanStartTime);
+        _extensionVoteEndTime = (_nextDueTime).add(_gracePeriod);
+        poolInfo[_pool].extensionVoteEndTime = _extensionVoteEndTime;
+        emit extensionRequested(_extensionVoteEndTime);
+    }
+
+    function voteOnExtension(address _pool) external isPoolActive {
```
ritik99, line 68: isPoolActive modifier has to be added. It also uses loanStatus and LoanStatus enum, so those might also have to be defined

PR #35, opened 2021-03-11
Reorg/repayments

ritik99:


contracts/Pool/Pool.sol
```diff
@@ -307,7 +272,7 @@ contract Pool is Initializable, IPool {
         uint256 _noOfRepaymentIntervals = poolConstants.noOfRepaymentIntervals;
         uint256 _repaymentInterval = poolConstants.repaymentInterval;
         IPoolFactory _poolFactory = IPoolFactory(PoolFactory);
-        IRepayment(_poolFactory.repaymentImpl()).initializeRepayment(_noOfRepaymentIntervals, _noOfRepaymentIntervals.mul(_repaymentInterval));
+        IRepayment(_poolFactory.repaymentImpl()).initializeRepayment(_noOfRepaymentIntervals, _repaymentInterval, poolConstants.borrowRate, poolConstants.loanStartTime);
```
prateekreddy, line 291: If these are not used in poolFactory,  then we can delete these variables.

contracts/Pool/Pool.sol
```diff
@@ -557,57 +503,35 @@ contract Pool is Initializable, IPool {
...
-            )
-        );
-        if (_interest < _extraInterest) {
-            _interest = 0;
-        } else {
-            _interest = _interest.sub(_extraInterest);
-        }
+        uint256 _repaymentPeriodCovered = IRepayment(_poolFactory.repaymentImpl()).getRepaymentPeriodCovered(address(this));
+        uint256 _interestAccruedThisPeriod = ((block.timestamp).sub(_repaymentPeriodCovered)).mul(_interestPerPeriod);
+        uint256 _repaymentOverdue = IRepayment(_poolFactory.repaymentImpl()).getRepaymentOverdue(address(this));
```
prateekreddy, line 549: Can we combine above call and this to a single fn call.

contracts/Pool/Pool.sol
prateekreddy: I meant poolConstants.

contracts/Pool/Pool.sol
prateekreddy: There are 2 instances of repayment contract calls na. Can we combine them into one ?

contracts/Pool/Pool.sol
ritik99: poolConstant vars are getting used in multiple places though, right?

prateekreddy: ohh... ok. Then it's fine.

contracts/Pool/Pool.sol
ritik99: Pushed new commit for this. getInterestCalculationVars() returns the two required variables

PR #39, opened 2021-03-18
Testing

ameesha1205:
Contract fixes

### DIFF
--- a/contracts/Pool/Pool.sol
+++ b/contracts/Pool/Pool.sol
@@ -197,7 +197,7 @@ contract Pool is Initializable, IPool {
     ) internal {
         //collateral ratio with 8 decimal precision
         uint256 price =IPriceOracle(IPoolFactory(PoolFactory).priceOracle()).getLatestPrice(poolConstants.borrowAsset, poolConstants.collateralAsset);
-        require(_amount >= poolConstants.idealCollateralRatio.mul(poolConstants.borrowAmountRequested.mul(price)).div(1e8), "36"); 
+        require(_amount >= poolConstants.idealCollateralRatio.mul(poolConstants.borrowAmountRequested.mul(price)).div(1e16), "36"); 
 
         _depositCollateral(_borrower, _amount, _transferFromSavingsAccount);
     }
--- a/contracts/Pool/Pool.sol
+++ b/contracts/Pool/Pool.sol
@@ -156,23 +156,25 @@ contract Pool is Initializable, IPool {
         uint256 _loanWithdrawalDuration,
         uint256 _collectionPeriod
     ) external payable initializer {
-        poolConstants.borrower = _borrower;
+        PoolFactory = msg.sender;
+        poolConstants.borrowAsset = _borrowAsset;
+        poolConstants.idealCollateralRatio = _idealCollateralRatio;
+        poolConstants.collateralAsset = _collateralAsset;
+        poolConstants.poolSavingsStrategy = _poolSavingsStrategy;
         poolConstants.borrowAmountRequested = _borrowAmountRequested;
+
+        _initialDeposit(_borrower, _collateralAmount, _transferFromSavingsAccount);
+
+        poolConstants.borrower = _borrower;
         poolConstants.minborrowAmount = _minborrowAmount;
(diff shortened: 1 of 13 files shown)


contracts/Pool/Pool.sol
```diff
@@ -182,36 +181,42 @@ contract Pool is Initializable, IPool {
...
     function depositCollateral(
+        address _borrower,
         uint256 _amount,
         bool _transferFromSavingsAccount
     ) public payable override {
         require(_amount != 0, "7");
+    
+        //collateral ratio with 8 decimal precision
+        uint256 price =IPriceOracle(IPoolFactory(PoolFactory).priceOracle()).getLatestPrice(poolConstants.borrowAsset, poolConstants.collateralAsset);
+        require(_amount >= poolConstants.idealCollateralRatio.mul(poolConstants.borrowAmountRequested.mul(price)).div(1e8), "36");
```
prateekreddy, line 192: Shouldn't this be in initialize function ? Otherwise we are adding additional restriction to depositCollateral right.

prateekreddy: All div should be 1e16. Because both price and idealCollateralRatio are in terms of e8

contracts/Pool/Pool.sol
```diff
@@ -182,36 +181,42 @@ contract Pool is Initializable, IPool {
...
         uint256 _sharesReceived =
             _depositToSavingsAccount(
                 _transferFromSavingsAccount,
                 poolConstants.collateralAsset,
                 _amount,
-                poolConstants.investedTo,
+                poolConstants.poolSavingsStrategy,
                 address(this),
-                msg.sender
+                _borrower
```
prateekreddy, line 202: This should probably be msg.sender only because borrower should not be able to use any random persons allowance na.

prateekreddy: We might have to revert probably to the older function only.

contracts/interfaces/IPool.sol
```diff
@@ -6,7 +6,7 @@ interface IPool {
     
     function getLoanStatus() external view returns (uint256);
 
-    function depositCollateral(uint256 _amount, bool _transferFromSavingsAccount) external payable;
+    function depositCollateral(address _borrower, uint256 _amount, bool _transferFromSavingsAccount) external payable;
```
prateekreddy, line 9: Based on the above comments in Pool, this also needs to be updated ?

contracts/mocks/Token.sol
```diff
@@ -2,13 +2,19 @@
...
+import "@openzeppelin/contracts/access/Ownable.sol";
 
-contract Token is ERC20 {
+contract Token is ERC20, Ownable {
     constructor(
         string memory name,
         string memory symbol,
         uint256 initialSupply
     ) ERC20(name, symbol) {
         _mint(msg.sender, initialSupply);
```
prateekreddy, line 13: Check if we need to initialize  ownable constructor as well

prateekreddy (changes requested):

contracts/Pool/Pool.sol
prateekreddy: Even more gas efficient probably is instead of multiplying with price, divide with inverse of price

prateekreddy: I was trying to fix this using 
```
function checkCollateral(uint256 _collateralAmount) internal view {
        uint256 _requiredCollateral = poolConstants.collateralRatio.mul(poolConstants.borrowAmountRequested).div(IPriceOracle(IPoolFactory(msg.sender).priceOracle()).getLatestPrice(poolConstants.collateralAsset, poolConstants.borrowAsset));
        require(_collateralAmount >= _requiredCollateral, "36");
    }
    ```

contracts/mocks/Token.sol
ameesha1205: Not needed

contracts/Pool/Pool.sol
prateekreddy: Can you update this please, so that I can merge.

prateekreddy (changes requested):

prateekreddy (approved):

PR #44, opened 2021-03-25
Fix/cancel pool

prateekreddy:
Cancel pool state management


prateekreddy:
Removed the deprecated deposit function from savingsAccount

prateekreddy:
Added ability to withdraw to a specified address rather than restricting to the sender only.

ritik99 (approved):

PR #48, opened 2021-04-01
Contract tests and updates

ameesha1205:


contracts/Pool/PoolFactory.sol
```diff
@@ -143,34 +164,146 @@ contract PoolFactory is Initializable, OwnableUpgradeable, IPoolFactory {
...
+                _transferFromSavingsAccount,
+                matchCollateralRatioInterval,
+                collectionPeriod
+            );
+
+        bytes32 salt = keccak256(abi.encodePacked(_salt, msg.sender));
+        bytes memory bytecode =
+            abi.encodePacked(
+                type(SublimeProxy).creationCode,
+                abi.encode(poolImpl, address(0x01), data)
```
prateekreddy, line 252: why is this address(0x01) ?

ameesha1205: This is updated from address(0) because the pool's view functions were reverting as the proxy do not allow admin to call the implementation.

prateekreddy: Can we probably view with non zero address ? But I guess this is also fine as we have to remember to do this everywhere which  will be a hassle.

ameesha1205: Yeah that's why

PR #56, opened 2021-05-24
Change Pool, Yield Contract

akshay111meher:
Confirm the changes in Aave Yield contract and Pool Contract before approving and merging the PR. 
@prateekreddy


contracts/SavingsAccount/SavingsAccount.sol
```diff
@@ -293,7 +293,10 @@ contract SavingsAccount is ISavingsAccount, Initializable, OwnableUpgradeable {
         }
 
         token = asset;
-        if (withdrawShares) token = IYield(strategy).liquidityToken(asset);
+        if (withdrawShares) {
+            token = IYield(strategy).liquidityToken(asset);
+            require(token != address(0), "Liquidity storage cannot be zero");
+        }
 
         if (token == address(0)) {
```
prateekreddy, line 301: This should be asset == address(0) because even if the token is Ether it is till invested in yeild right, but if strategy is address(0), then the tokens are held in the   savings  contract itself

contracts/yield/AaveYield.sol
```diff
@@ -201,6 +201,20 @@ contract AaveYield is IYield, Initializable, OwnableUpgradeable {
         emit UnlockedTokens(asset, received);
     }
 
+    function unlockShares(address asset, uint256 amount)
+        public
+        override
+        onlySavingsAccount
+        returns (uint256 received)
+    {
+        require(amount != 0, "Invest: amount");
```
prateekreddy, line 210: change this  to if condition rather than require

contracts/yield/CompoundYield.sol
```diff
@@ -139,6 +139,20 @@ contract CompoundYield is IYield, Initializable, OwnableUpgradeable {
         emit UnlockedTokens(asset, received);
     }
 
+    function unlockShares(address asset, uint256 amount)
+        public
+        override
+        onlySavingsAccount
+        returns (uint256 received)
+    {
+        require(amount != 0, "Invest: amount");
```
prateekreddy, line 148: change this to if condition rather than require

contracts/yield/YearnYield.sol
```diff
@@ -139,6 +139,20 @@ contract YearnYield is IYield, Initializable, OwnableUpgradeable {
         emit UnlockedTokens(asset, received);
     }
 
+    function unlockShares(address asset, uint256 amount)
+        public
+        override
+        onlySavingsAccount
+        returns (uint256 received)
+    {
+        require(amount != 0, "Invest: amount");
```
prateekreddy, line 148: change this to if condition rather than require

contracts/SavingsAccount/SavingsAccount.sol
akshay111meher: The withdraw function is re-written now!.

contracts/yield/AaveYield.sol
akshay111meher: Done!.

contracts/yield/CompoundYield.sol
akshay111meher: Done!.

contracts/yield/YearnYield.sol
akshay111meher: Done!.

prateekreddy (approved):

PR #57, opened 2021-05-25
Fix: Can burn when paused

prateekreddy:
In  the current contracts, pool token is paused when pool is cancelled so lender cannot withdraw the liquidity by burning tokens as burn and mint are also disabled when token is paused. This PR makes the burn possible even when pool token is paused.


ritik99:
Looks good

PR #58, opened 2021-05-26
Fix: before pool token transfer margincall status

prateekreddy:
Pool token transfer is only possible if from and to addresses haven't invoked a margin call. In the current code, the check on margin call status was opposite. This PR fixes the issue.


ritik99:
Looks good

PR #59, opened 2021-06-08
cancelling pool incurs penalty

prateekreddy:


contracts/Pool/Pool.sol
```diff
@@ -570,6 +610,10 @@ contract Pool is Initializable, IPool, ReentrancyGuard {
             _balance = _balance.mul(_totalAsset).div(poolToken.totalSupply());
         }
 
+        if(_loanStatus == LoanStatus.CANCELLED) {
+            _balance = _balance.mul(poolVars.penalityLiquidityAmount).div(poolToken.totalSupply());
```
ritik99, line 614: Shouldn't this be _balance = _balance(1 + penalityLiquidityAmount / totalSupply)?

contracts/Pool/Pool.sol
```diff
@@ -570,6 +610,10 @@ contract Pool is Initializable, IPool, ReentrancyGuard {
             _balance = _balance.mul(_totalAsset).div(poolToken.totalSupply());
```
ritik99, line 610: Are we handling the case of withdrawing liquidity after lender margin called separately? Might need separate logic if lender withdrawing liquidity has margin called (and then the borrower is getting liquidated) vs when the borrower defaults by missing repayment

ritik99 (changes requested):
Added two comments

contracts/Pool/Pool.sol
prateekreddy: correct, fixing it.

contracts/Pool/Pool.sol
prateekreddy: When lender margin calls, liquidator directly transfers the lent tokens to lender. So the balance of the lender will be 0.

ritik99:
Looks good!

ritik99 (approved):

PR #60, opened 2021-06-09
Lending added

akshay111meher:


contracts/Pool/Pool.sol
```diff
@@ -16,6 +14,8 @@ import "../interfaces/IPool.sol";
 import "../interfaces/IExtension.sol";
 import "../interfaces/IPoolToken.sol";
 
+import "hardhat/console.sol";
```
prateekreddy, line 17: Can you remove this ?

contracts/Pool/Pool.sol
```diff
@@ -275,12 +275,19 @@ contract Pool is Initializable, IPool, ReentrancyGuard {
...
+                //     _poolSavingsStrategy,
+                //     _liquidityshare
+                // );
                 _sharesReceived = _savingsAccount.transferFrom(
                     _asset,
                     _depositFrom,
                     address(this),
                     _poolSavingsStrategy,
-                    _liquidityshare
+                    _amount
```
prateekreddy, line 290: Why this change ?

contracts/Pool/Pool.sol
```diff
@@ -275,12 +275,19 @@ contract Pool is Initializable, IPool, ReentrancyGuard {
                         _asset
                     );
             if (_toSavingsAccount) {
+                // _sharesReceived = _savingsAccount.transferFrom(
+                //     _asset,
+                //     _depositFrom,
+                //     address(this),
+                //     _poolSavingsStrategy,
+                //     _liquidityshare
+                // );
```
prateekreddy, line 284: Don't comment, remove completely.

prateekreddy (changes requested):

contracts/Pool/Pool.sol
akshay111meher: liquidity shares seem to be a wrong input there. 

When consoled, _amount was making the transaction pass. 

Tried simulating with some tokens as well.

contracts/Pool/Pool.sol
akshay111meher: Will resolve this once, the above comment is confirmed..

contracts/Pool/Pool.sol
prateekreddy: Makes sense, savings account should  take actually value as input, not the liquidity shares

akshay111meher:
The merge is redundant. 
Inclusive changes in other branch

PR #64, opened 2021-06-20
Fix/savings account amount

prateekreddy:


contracts/CreditLine/CreditLine.sol
```diff
@@ -482,7 +478,7 @@ contract CreditLine is CreditLineStorage, ReentrancyGuard {
                 _savingsAccount.withdrawFrom(
                     _lender,
                     address(this),
-                    _liquidityShares,
+                    _tokensToTransfer,
                     _asset,
                     _strategyList[_index],
                     false
```
ritik99, line 484: Can be put within the else block

ritik99 (changes requested):
one suggestion regarding the moving the withdrawFrom call to within the else block

contracts/CreditLine/CreditLine.sol
prateekreddy: done

ritik99 (approved):
Looks good

PR #69, opened 2021-06-22
Test: Pool Withdraw deadline phase

prateekreddy:


contracts/Pool/Pool.sol
```diff
@@ -489,28 +493,23 @@ contract Pool is Initializable, IPool, ReentrancyGuard {
...
-        _withdrawFromSavingsAccount(
-            ISavingsAccount(IPoolFactory(PoolFactory).savingsAccount()), 
-            address(this), 
-            msg.sender, 
-            _tokensLent, 
-            poolConstants.borrowAsset, 
-            address(0), 
-            false
-        );
+        IERC20(poolConstants.borrowAsset).transfer(msg.sender, _tokensLent);
```
ritik99, line 496: made this change because the assumption is that during the collection period the collected amount stays within the pool contract only na?

ritik99 (approved):

contracts/Pool/Pool.sol
prateekreddy: yeah....

PR #74, opened 2021-07-02
Fix/repayments

ritik99:
Changes surrounding repayments. Largely:
1. Switched to a single repayAmount() function
2. Added function to check if borrower defaulted
3. Some name changes


ritik99:
There's a couple of TODOs still marked.

Major changes:
1. Fixed base to 10**30 from 10**8 where I could find
2. periodWhenExtensionIsPassed in Extension.sol now stores the period for which thee extension was passed
3. updating and keeping track of nextDuePeriod in Pool.sol and Repayments.sol is probably not necessary anymore since I've added a view function getNextInstalmentDeadline() in Repayments.sol
4. Separated RepaymentDetails into RepaymentConstants and RepaymentVars
5. Added function to check if borrower has defaulted didBorrowerDefault()
6. Made changes to repayAmoutn() and added some helpers for it


Some cleaning might be required for unused functions/vars

ritik99:
1. Merged changes
2. Removed repaymentsOverdue
3. Added function getInterestDueTillInstalmentDeadline() to get interest due till the next deadline
4. Changed the modifier for closeLoan() in Pool.sol
5. Changed implementation of interestTillNow() in Pool.sol to make it simpler

PR #82, opened 2021-07-22
Audit fixes

prateekreddy:
Audit findings from Peckshield


contracts/Repayments/Repayments.sol
```diff
@@ -407,7 +407,7 @@ contract Repayments is Initializable, RepaymentStorage, IRepayment, ReentrancyGu
     }
     */
     function getTotalRepaidAmount(address poolID) external view override returns (uint256) {
-        return repaymentVars[poolID].totalRepaidAmount;
+        return repaymentVars[poolID].loanDurationCovered.mul(getInterestPerSecond(_poolID)).div(10**60);
```
ritik99: could margin calls cause an issue here? For eg, borrower repays some amount, but then after that the lender liquidates via margin call, so then the getInterestPerSecond() function calculates lesser amount than was actually repaid by the borrower

contracts/Repayments/Repayments.sol
```diff
@@ -407,7 +407,7 @@ contract Repayments is Initializable, RepaymentStorage, IRepayment, ReentrancyGu
     }
     */
     function getTotalRepaidAmount(address poolID) external view override returns (uint256) {
-        return repaymentVars[poolID].loanDurationCovered.mul(getInterestPerSecond(_poolID)).div(10**60);
+        return repaymentVars[poolID].loanDurationCovered.mul(getInterestPerSecond(poolID)).div(10**60);
```
ritik99, line 364: args will have an underscore acc. to the convention we're following na

ritik99 (commented):
two other minor comments: savingsAccount and penalty & savingAccount and penality in some places. Let's just stick to the first convention I guess

contracts/Repayments/Repayments.sol
prateekreddy: Right, makes sense. Do you think we should maintain a different variable for this ?

prateekreddy: Added a repaidAmount variable to track repayments to a pool

PR #89, opened 2021-08-02
Feature/volatility per token

prateekreddy:


contracts/Pool/PoolFactory.sol
```diff
@@ -266,7 +265,6 @@ contract PoolFactory is Initializable, OwnableUpgradeable, IPoolFactory {
         _updateCollectionPeriod(_collectionPeriod);
         _updateMatchCollateralRatioInterval(_matchCollateralRatioInterval);
         _updateMarginCallDuration(_marginCallDuration);
-        _updateCollateralVolatilityThreshold(_collateralVolatilityThreshold);
```
ritik99, line 269: should we be initializing it in setImplementations()?

ritik99 (approved):

PR #99, opened 2021-09-09
Add verification tests

akshay111meher:
No change in contract. May be in some places linter-has chagned!..


akshay111meher:
The verification contracts were changed. Hence the tests in this branch are irrelevant

PR #101, opened 2021-09-17
Misc improvements

prateekreddy:


ritik99 (changes requested):
Apart from the name changes for repayments state, info and all I think rest looks good

PR #103, opened 2021-09-28
No strategy

akshay111meher:
No Strategy yield is defined. avoid use of zeroAddress in contracts


contracts/Pool/PoolFactory.sol
```diff
@@ -890,6 +892,10 @@ contract PoolFactory is Initializable, OwnableUpgradeable, IPoolFactory {
         emit LimitsUpdated('NoOfRepaymentIntervals', _min, _max);
     }
 
+    function updateNoYield(address _noYield) external onlyOwner {
+        noYield = _noYield;
```
prateekreddy, line 896: Event needs to be emitted right. Can we follow the same pattern as other setters, one internal and one external function interval will be used in initialize and external will be only owner

contracts/SavingsAccount/SavingsAccount.sol
```diff
@@ -135,29 +122,23 @@ contract SavingsAccount is ISavingsAccount, Initializable, OwnableUpgradeable, R
...
         userLockedBalance[msg.sender][asset][currentStrategy] = userLockedBalance[msg.sender][asset][currentStrategy].sub(
             amount,
             'SavingsAccount::switchStrategy Insufficient balance'
         );
 
         uint256 tokensReceived = amount;
-        if (currentStrategy != address(0)) {
-            tokensReceived = IYield(currentStrategy).unlockTokens(asset, amount);
-        }
+        tokensReceived = IYield(currentStrategy).unlockTokens(asset, amount);
```
prateekreddy, line 137: We can probably directly set tokensReceived while initializing, no need to set it to amount right and then chagne right

contracts/SavingsAccount/SavingsAccount.sol
```diff
@@ -135,29 +122,23 @@ contract SavingsAccount is ISavingsAccount, Initializable, OwnableUpgradeable, R
...
 
         uint256 sharesReceived = tokensReceived;
-        if (newStrategy != address(0)) {
-            if (asset != address(0)) {
-                IERC20(asset).safeApprove(newStrategy, tokensReceived);
-            }
-
-            sharesReceived = _depositToYield(tokensReceived, asset, newStrategy);
-        }
+        IERC20(asset).safeApprove(newStrategy, tokensReceived);
```
prateekreddy, line 140: We can probably directly set sharesReceived while initializing, no need to set it to amount right and then chagne right. Please take care of this everywhere else as well

contracts/yield/CompoundYield.sol
```diff
@@ -51,22 +57,19 @@ contract CompoundYield is IYield, Initializable, OwnableUpgradeable, ReentrancyG
     }
 
     function updateProtocolAddresses(address _asset, address _to) external onlyOwner {
+        require(_asset != address(0), 'asset cannot be address(0)');
         liquidityToken[_asset] = _to;
         emit ProtocolAddressesUpdated(_asset, _to);
     }
 
     function emergencyWithdraw(address _asset, address payable _wallet) external onlyOwner returns (uint256 received) {
+        require(_asset != address(0), 'asset cannot be address(0)');
```
prateekreddy, line 66: May be, let's not do this, just in case

contracts/yield/CompoundYield.sol
```diff
@@ -83,12 +86,14 @@ contract CompoundYield is IYield, Initializable, OwnableUpgradeable, ReentrancyG
...
     ) public payable override onlySavingsAccount nonReentrant returns (uint256 sharesReceived) {
         require(amount != 0, 'Invest: amount');
+        require(asset != address(0), 'asset cannot be address(0)');
+
         address investedTo = liquidityToken[asset];
-        if (asset == address(0)) {
-            require(msg.value == amount, 'Invest: ETH amount');
+        IERC20(asset).safeTransferFrom(user, address(this), amount);
+        if (asset == address(iweth)) {
+            iweth.withdraw(amount);
```
prateekreddy, line 94: Can't we directly deposit WETH ?

contracts/yield/CompoundYield.sol
```diff
@@ -102,26 +107,28 @@ contract CompoundYield is IYield, Initializable, OwnableUpgradeable, ReentrancyG
      **/
     function unlockTokens(address asset, uint256 amount) public override onlySavingsAccount nonReentrant returns (uint256 received) {
         require(amount != 0, 'Invest: amount');
+        require(asset != address(0), 'asset cannot be address(0)');
+
         address investedTo = liquidityToken[asset];
 
-        if (asset == address(0)) {
+        if (asset == address(iweth)) {
```
prateekreddy, line 114: Can't we directly deposit WETH ?

contracts/yield/YearnYield.sol
```diff
@@ -39,6 +41,11 @@ contract YearnYield is IYield, Initializable, OwnableUpgradeable, ReentrancyGuar
         _updateSavingsAccount(_savingsAccount);
     }
 
+    function updateIweth9(address payable _iwethAddress) public onlyOwner {
+        require(_iwethAddress != address(0), '_iwethAddress address cannot be zero');
```
prateekreddy, line 45: emit event when updated. Follow the internal for setting within contract and external for owner pattern

contracts/yield/YearnYield.sol
```diff
@@ -50,22 +57,19 @@ contract YearnYield is IYield, Initializable, OwnableUpgradeable, ReentrancyGuar
     }
 
     function updateProtocolAddresses(address asset, address to) external onlyOwner {
+        require(asset != address(0), 'asset cannot be address(0)');
         liquidityToken[asset] = to;
         emit ProtocolAddressesUpdated(asset, to);
     }
 
     function emergencyWithdraw(address _asset, address payable _wallet) external onlyOwner nonReentrant returns (uint256 received) {
+        require(_asset != address(0), 'asset cannot be address(0)');
```
prateekreddy, line 66: Let's not do this, just in case

contracts/yield/YearnYield.sol
```diff
@@ -82,16 +86,16 @@ contract YearnYield is IYield, Initializable, OwnableUpgradeable, ReentrancyGuar
...
     ) public payable override onlySavingsAccount nonReentrant returns (uint256 sharesReceived) {
         require(amount != 0, 'Invest: amount');
+        require(asset != address(0), 'asset cannot be address(0)');
 
         address investedTo = liquidityToken[asset];
-        if (asset == address(0)) {
-            require(msg.value == amount, 'Invest: ETH amount');
+        IERC20(asset).safeTransferFrom(user, address(this), amount);
+        if (asset == address(iweth)) {
+            iweth.withdraw(amount);
```
prateekreddy, line 94: Can't we directly deposit WETH ?

prateekreddy (changes requested):

contracts/Pool/PoolFactory.sol
akshay111meher: Will add an event, 
was actually planning to not disturb the `initialize` function, hence this

contracts/yield/CompoundYield.sol
akshay111meher: I got your concern, 
better write a function named `emergencyWithdrawEth` rather than modifying it

contracts/yield/CompoundYield.sol
akshay111meher: No, Compound doesn't know anything of WETH. It has different contract set for eth and token.

Uniswap, Aave, Yearn use WETH, (sadly WETH used by Yearn and Aave is different, hence we code all these in yield contracts)

contracts/yield/CompoundYield.sol
akshay111meher: No, Compound doesn't know anything of WETH. It has different contract set for eth and token.

Uniswap, Aave, Yearn use WETH, (sadly WETH used by Yearn and Aave is different, hence we code all these in yield contracts)

contracts/yield/YearnYield.sol
akshay111meher: No!.
Clearer discussion on this can be done on discord!.

contracts/yield/CompoundYield.sol
prateekreddy: why ?

contracts/yield/CompoundYield.sol
prateekreddy: Ahh... ok.

contracts/yield/CompoundYield.sol
akshay111meher: Just a cleaner code!...
don't want many nested if else conditions..

akshay111meher:
The PR is closed as the feature is implemented in other branch.

PR #117, opened 2021-10-26
Credit lines strategy priority

akshay111meher:
Amount sent from lender to borrower is withdrawn in the order of strateiges provided by lender.


prateekreddy:
This feature is deferred keeping in view the code freeze due to audit.

PR #121, opened 2021-11-01
Fix/extension voting

prateekreddy:


prateekreddy:
One of the issues with current implementation is that for every token transfer, couple of inter contract calls to be made. One way to decrease the gas costs is to store extension address in pool contract itself instead of querying  from poolFactory. But this reduces our ability to change extension contract code if needed for existing pools.

PR #123, opened 2021-11-03
Adding ETH as collateral and borrow token

KalpitaMandal:
Adding test cases for creditlines with ETH as borrow and collateral token


ritik99:
Approvals seem to be the only contract changes, the rest are all changes in tests. Seems fine.

prateekreddy:
ETH is no longer being used as collateral/borrow token instead WETH is being used

PR #135, opened 2021-11-16
Updated the AmountBorrowed event

saxenism:
Updated the `AmountBorrowed` event (https://github.com/sublime-finance/sublime-v1/blob/update-borrow-event/contracts/Pool/Pool.sol#L342) to have an extra parameter of `_protocolFee`.

### DIFF
--- a/contracts/Pool/Pool.sol
+++ b/contracts/Pool/Pool.sol
@@ -339,7 +339,7 @@ contract Pool is Initializable, ERC20PausableUpgradeable, IPool, ReentrancyGuard
         SavingsAccountUtil.transferTokens(_borrowAsset, _protocolFee, address(this), _collector);
         SavingsAccountUtil.transferTokens(_borrowAsset, _tokensLent.sub(_protocolFee), address(this), msg.sender);
 
-        emit AmountBorrowed(_tokensLent, _protocolFee);
+        emit AmountBorrowed(_tokensLent.sub(_protocolFee), _protocolFee);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


contracts/Pool/Pool.sol
```diff
@@ -339,7 +339,7 @@ contract Pool is Initializable, ERC20PausableUpgradeable, IPool, ReentrancyGuard
         SavingsAccountUtil.transferTokens(_borrowAsset, _protocolFee, address(this), _collector);
         SavingsAccountUtil.transferTokens(_borrowAsset, _tokensLent.sub(_protocolFee), address(this), msg.sender);
 
-        emit AmountBorrowed(_tokensLent);
+        emit AmountBorrowed(_tokensLent, _protocolFee);
```
prateekreddy, line 342: let's use _tokensLent.sub(_protocolFee) as the first arg

prateekreddy (changes requested):

contracts/Pool/Pool.sol
saxenism: Yeah, that makes more sense. Will update.

contracts/Pool/Pool.sol
```diff
@@ -339,7 +339,7 @@ contract Pool is Initializable, ERC20PausableUpgradeable, IPool, ReentrancyGuard
         SavingsAccountUtil.transferTokens(_borrowAsset, _protocolFee, address(this), _collector);
         SavingsAccountUtil.transferTokens(_borrowAsset, _tokensLent.sub(_protocolFee), address(this), msg.sender);
 
-        emit AmountBorrowed(_tokensLent);
+        emit AmountBorrowed(_tokensLent.sub(_protocolFee), _protocolFee);
```
prateekreddy, line 342: Lets not do the sub 2 times, we already do that above right, instead may be we can assign to a memory variable and reuse.

prateekreddy (changes requested):

prateekreddy (approved):

PR #138, opened 2021-11-17
fix replay attack while linking to master address

prateekreddy:


prateekreddy:
Not relevant after #142

PR #160, opened 2021-12-27
Lime 69, Issue 5, Gas saving by duplicate check

saxenism:
## Description

`maxStrategies` check needs to be added to `updateMaxStrategies` instead of `_updateMaxStrategies`, to save gas by decreasing the code size. 

PR to fix issue listed at: https://github.com/code-423n4/2021-12-sublime-findings/issues/5

## Integration Checklist

- [ ] Compare gas savings from gasReport.md

## Changelog

Removed duplicate checks for *_maxStrategies* and optimized gas usage.

### DIFF
--- a/contracts/yield/StrategyRegistry.sol
+++ b/contracts/yield/StrategyRegistry.sol
@@ -31,7 +31,6 @@ contract StrategyRegistry is Initializable, OwnableUpgradeable, IStrategyRegistr
      * @param _maxStrategies maximum number of strategies allowed
      **/
     function initialize(address _owner, uint256 _maxStrategies) external initializer {
-        require(_maxStrategies != 0, 'StrategyRegistry::initialize maxStrategies cannot be zero');
         __Ownable_init();
         super.transferOwnership(_owner);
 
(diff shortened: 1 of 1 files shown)


saxenism:
I've added the require statement to the external function instead of the internal function as stated in the Lime 69 issue. But I was wondering, whether the require statement should have the condition **_maxStrategies > 0** instead of **_maxStrategies != 0** to filter out negative values of _maxStrategies.

@prateekreddy @ritik99

prateekreddy:
> condition _maxStrategies > 0 instead of _maxStrategies != 0 to filter out negative values of _maxStrategies.

Max strategies cannot be negative because it is unsigned integer

prateekreddy (changes requested):
Instead of moving maxStrategies check, we can remove the require statement in initialize as suggested.

saxenism:
> Max strategies cannot be negative because it is unsigned integer

Oh lol, yes true.

> Instead of moving maxStrategies check, we can remove the require statement in initialize as suggested.

Yes, agreed. Made the changes. Please review again.

PR #163, opened 2021-12-27
Lime 75, Issue 36, Removed default var initialization

saxenism:
## Description:
uint256 variable are initialized to a default value of zero per Solidity docs. Setting a variable to the default value is unnecessary. Removing lines of code where variables are initialized to zero can save gas.

PR made to fix issue at: https://github.com/code-423n4/2021-12-sublime-findings/issues/36

## Integrations Checklist
- [ ] Compare gas savings from gasReport.md

## Changelog
Throughout the contracts, all variables that were being initialized to their default values only, have been changed to getting only declared.


prateekreddy:
@saxenism For all the loops also index is being initialized to 0

prateekreddy (changes requested):
@saxenism For all the loops also index is being initialized to 0

saxenism:
> @saxenism For all the loops also index is being initialized to 0

Good catch. Made the changes. Please review again.

PR #164, opened 2021-12-27
Lime 76, Issue 38, Named return issues resolved

saxenism:
## Description
If a function returns something and in the function signature, the name of the returning variable along with the type is mentioned, it is a good practice to use that name while returning a value from the function. 
 
PR opened to fix issue at: https://github.com/code-423n4/2021-12-sublime-findings/issues/38

## Change Log
With this edit of named return we implemented an improvement in code styling/clarity.


prateekreddy:
Has this style been consistent across all functions ?

prateekreddy (changes requested):
@saxenism Please check if this style been consistent across all functions ? Let's ensure we use return everywhere even in case of multiple return value

saxenism:
> @saxenism Please check if this style been consistent across all functions ? Let's ensure we use return everywhere even in case of multiple return value

Understood. Switched the return style. Please review again.

ritik99 (commented):
@saxenism I don't think the styling is still consistent throughout the contracts. Some instances: [1](https://github.com/sublime-finance/sublime-v1/blob/f25860e182e4c9a8fed0332f1cbd4755d68b3af4/contracts/Pool/Pool.sol#L244), [2](https://github.com/sublime-finance/sublime-v1/blob/f25860e182e4c9a8fed0332f1cbd4755d68b3af4/contracts/Pool/Pool.sol#L165). Please recheck all the contracts

saxenism:
> @saxenism I don't think the styling is still consistent throughout the contracts. Some instances: [1](https://github.com/sublime-finance/sublime-v1/blob/f25860e182e4c9a8fed0332f1cbd4755d68b3af4/contracts/Pool/Pool.sol#L244), [2](https://github.com/sublime-finance/sublime-v1/blob/f25860e182e4c9a8fed0332f1cbd4755d68b3af4/contracts/Pool/Pool.sol#L165). Please recheck all the contracts

Scanned all contracts and made the return style consistent. Please check again.

Few functions that I did not change:
1. [SavingsAccount.withdrawAll](https://github.com/sublime-finance/sublime-v1/blob/master/contracts/SavingsAccount/SavingsAccount.sol#L304)

This is because I did not see the function returning at all. Maybe we can discuss the function's return declaration. 

2. [PoolFactory._deploy](https://github.com/sublime-finance/sublime-v1/blob/master/contracts/Pool/PoolFactory.sol#L404)

This is because I didn't really want to change a function making an assembly call.

3.[ NoYield.emergencyWithdraw](https://github.com/sublime-finance/sublime-v1/blob/master/contracts/yield/NoYield.sol#L78)

This is because the function's logic is to be changed from another PR, and style changes can be made after that.

prateekreddy:
@saxenism IsavingsAccount has some inconsistency with the format. Can you check ?

prateekreddy:
In all the interfaces let's name the return values, so it is more clear. We need not do this on actual functions though.

saxenism:
> @saxenism In all the interfaces let's name the return values, so it is more clear. We need not do this on actual functions though.
> 
> PoolFactory, SavingsAccount, NoYield has inconsistent return patterns. Please check.

Made the required changes, apart from `NoYield.emergencyWithdraw` because that function has yet to be changed.
Please review again.

contracts/SavingsAccount/SavingsAccountUtil.sol
```diff
@@ -47,11 +47,10 @@ library SavingsAccountUtil {
         address _strategy,
         address _from,
         address _to,
-        uint256 _amount,
-        address _token,
-        address _strategy
+        uint256 _amount
     ) internal returns (uint256) {
-        transferTokens(_token, _amount, _from, address(this));
+        uint256 _sharesReceived;
```
prateekreddy, line 52: This can be declared at initialization ?

contracts/interfaces/IPool.sol
```diff
@@ -94,13 +94,13 @@ interface IPool {
 
     function withdrawBorrowedAmount() external;
 
-    function borrower() external returns (address);
+    function borrower() external returns (address poolBorrower);
 
-    function getMarginCallEndTime(address _lender) external returns (uint256);
+    function getMarginCallEndTime(address _lender) external returns (uint256 marginCallEndTimeFor_lender);
```
prateekreddy, line 99: Inconsistent naming convention

contracts/interfaces/IPriceOracle.sol
```diff
@@ -24,7 +24,7 @@ interface IPriceOracle {
      */
     event UniswapPriceAveragingPeriodUpdated(uint32 uniswapPriceAveragingPeriod);
 
-    function getLatestPrice(address num, address den) external view returns (uint256, uint256);
+    function getLatestPrice(address num, address den) external view returns (uint256 priceOfNumInTermsOfDen, uint256 priceDecimals);
```
prateekreddy, line 27: (uint256 price, uint256 decimals) should be sufficient ?

contracts/interfaces/IStrategyRegistry.sol
```diff
@@ -20,9 +20,9 @@ interface IStrategyRegistry {
      */
     event MaxStrategiesUpdated(uint256 maxStrategies);
 
-    function registry(address _strategy) external view returns (bool);
+    function registry(address _strategy) external view returns (bool isValidStrategy);
 
-    function getStrategies() external view returns (address[] memory);
+    function getStrategies() external view returns (address[] memory strategiesArray);
```
prateekreddy, line 25: strategies should be sufficient

contracts/yield/NoYield.sol
```diff
@@ -48,7 +48,7 @@ contract NoYield is IYield, Initializable, OwnableUpgradeable, ReentrancyGuard {
     /**
      * @notice used to query liquidity token for a given asset
      * @param _asset address of the asset
-     * @return _tokenAddress: address of the lqiudity token for the asset
+     * @return _asset: address of the lqiudity token for the asset
```
prateekreddy, line 51: return value and param have same var name

prateekreddy (changes requested):

contracts/yield/NoYield.sol
```diff
@@ -48,10 +48,11 @@ contract NoYield is IYield, Initializable, OwnableUpgradeable, ReentrancyGuard {
     /**
      * @notice used to query liquidity token for a given asset
      * @param _asset address of the asset
-     * @return _asset: address of the lqiudity token for the asset
+     * @return _tokenAddress: address of the lqiudity token for the asset
      **/
     function liquidityToken(address _asset) external pure override returns (address) {
-        return _asset;
+        address _tokenAddress = _asset;
```
prateekreddy, line 54: you can return asset directly, assigning to tokenAddress is not necessary

prateekreddy (changes requested):

prateekreddy (approved):

PR #168, opened 2021-12-27
Lime 98, Issue 82, Storing 30 as a constant

saxenism:
## Description
We use 30 as the power to which numbers are raised to save precision. Currently this integer value is used directly without a clear indication that this value relates to the decimals value, which could lead to one of these values being modified but not the other (perhaps by a typo), which is the basis for many past hacks. Coding best practices suggests using a constant integer to store this value in a way that clearly explains the purpose of this value to prevent confusion.

PR opened to fix issue at: https://github.com/code-423n4/2021-12-sublime-findings/issues/82

## Integration Checklist

- [ ] Decide upon the correct name of this constant and update documentations

## Change Log
Documentation improvement

## Note to Reviewer
For some reason, line 119 in `contracts/PriceOracle.sol` (check Evading the error commit) using the SCALING_FACTOR (1e30) constant always threw an overflow error.


saxenism:
For some reason, line 119 in `contracts/PriceOracle.sol` (check *Evading the error* commit) using the SCALING_FACTOR (1e30) constant always threw an overflow error.

contracts/Pool/PoolFactory.sol
```diff
@@ -96,6 +96,8 @@ contract PoolFactory is Initializable, OwnableUpgradeable, IPoolFactory {
      */
     uint256 public override poolCancelPenaltyMultiple;
 
+    uint256 constant SCALING_FACTOR = 1e30;
```
prateekreddy, line 99: This need not be declared as we aren't using the variable anywhere

contracts/PriceOracle.sol
```diff
@@ -73,12 +74,12 @@ contract PriceOracle is Initializable, OwnableUpgradeable, IPriceOracle {
...
         uint256 price = uint256(price1)
             .mul(10**_feedData2.decimals)
-            .mul(10**30)
+            .mul(SCALING_FACTOR)
             .div(uint256(price2))
             .div(10**_feedData1.decimals)
             .mul(10**decimals[den])
             .div(10**decimals[num]);
-        return (price, 30);
+        return (price, SCALING_FACTOR);
```
prateekreddy, line 82: 30 shouldn't be replaced by 1e30

contracts/PriceOracle.sol
```diff
@@ -115,8 +116,8 @@ contract PriceOracle is Initializable, OwnableUpgradeable, IPriceOracle {
         }
 
         int24 _twapTick = OracleLibrary.consult(_pool, uniswapPriceAveragingPeriod);
-        uint256 _numTokens = OracleLibrary.getQuoteAtTick(_twapTick, 10**30, num, den);
-        return (_numTokens, 30);
+        uint256 _numTokens = OracleLibrary.getQuoteAtTick(_twapTick, 1e30, num, den);
+        return (_numTokens, SCALING_FACTOR);
```
prateekreddy, line 120: 30 shouldn't be replaced by 1e30

contracts/PriceOracle.sol
```diff
@@ -115,8 +116,8 @@ contract PriceOracle is Initializable, OwnableUpgradeable, IPriceOracle {
         }
 
         int24 _twapTick = OracleLibrary.consult(_pool, uniswapPriceAveragingPeriod);
-        uint256 _numTokens = OracleLibrary.getQuoteAtTick(_twapTick, 10**30, num, den);
-        return (_numTokens, 30);
+        uint256 _numTokens = OracleLibrary.getQuoteAtTick(_twapTick, 1e30, num, den);
```
prateekreddy, line 119: Use `SCALING_FACTOR` instead of `1e30`

prateekreddy (changes requested):

contracts/PriceOracle.sol
saxenism: Done

contracts/PriceOracle.sol
saxenism: Done

contracts/PriceOracle.sol
saxenism: Done

contracts/Pool/PoolFactory.sol
saxenism: True, it was only being used for explanations in Natspec comments. Removed this declaration and changed the Natspec docs from using the SCALING_FACTOR to 10**30

prateekreddy: hmm instead probably use `SCALING_FACTOR(10**30)` ? That would probably be more clear

contracts/PriceOracle.sol
```diff
@@ -13,6 +13,9 @@ contract PriceOracle is Initializable, OwnableUpgradeable, IPriceOracle {
     using SafeMath for uint256;
 
     uint32 uniswapPriceAveragingPeriod;
+    uint256 constant SCALING_FACTOR = 1e30;
```
prateekreddy, line 16: Use 10**(SCALING_NUMBER) instead

prateekreddy (changes requested):

contracts/Pool/PoolFactory.sol
saxenism: Yes, agreed. That would definitely be more clear, but I wanted to incorporate the suggestions over at https://github.com/code-423n4/2021-12-sublime-findings/issues/126 

Let me know if you think we should ignore this.

contracts/PriceOracle.sol
saxenism: We can do that. But then it would not be in line with this https://github.com/code-423n4/2021-12-sublime-findings/issues/126

Let me know if you think we can ignore this.

contracts/Pool/PoolFactory.sol
saxenism: Done. Please review again.

contracts/PriceOracle.sol
saxenism: Done. Please review again.

contracts/Pool/PoolFactory.sol
```diff
@@ -737,7 +737,7 @@ contract PoolFactory is Initializable, OwnableUpgradeable, IPoolFactory {
 
     /**
      * @notice used to query protocol fee fraction and address of the collector
-     * @return protocolFee Fraction multiplied by 10**30
+     * @return protocolFee Fraction multiplied by SCALING_FACTOR(SCALING_FACTOR(10**30))
```
ritik99, line 740: extra SCALING_FACTOR

contracts/PriceOracle.sol
```diff
@@ -13,6 +13,9 @@ contract PriceOracle is Initializable, OwnableUpgradeable, IPriceOracle {
     using SafeMath for uint256;
 
     uint32 uniswapPriceAveragingPeriod;
+    uint256 constant SCALING_NUMBER = 30;
```
ritik99, line 16: Rename SCALING_NUMBER to SCALING_EXPONENT

ritik99 (commented):
Check [line 675](https://github.com/sublime-finance/sublime-v1/blob/b1718b676dcdda936309f04ed8bb18e097bcf8cd/contracts/Pool/Pool.sol#L675), it could be changed to SCALING_FACTOR**2

contracts/PriceOracle.sol
saxenism: Done.

contracts/Pool/PoolFactory.sol
saxenism: Yep. Removed.

PR #180, opened 2021-12-27
Lime-86, Issue 65: PoolConstants packing

KalpitaMandal:
# Description
Stacked pull request on #178
In the `PoolConstants` struct we have a number of fields which don't need to take up an entire storage slot. Easy examples are the timestamps which can be placed in a `uint64` variable and packed with an address.

https://github.com/code-423n4/2021-12-sublime/blob/9df1b7c4247f8631647c7627a8da9bdc16db8b11/contracts/Pool/Pool.sol#L46-L59

This will save gas when performing the initial writes of these structs.

# Integrations Checklist

- [ ]  Have any function signatures changed? If yes, outline below.
- [ ]  Have any features changed or been added? If yes, outline below.
- [ ]  Have any events changed or been added? If yes, outline below.
- [ ]  Has all documentation been updated?

# Changelog

- Restructured the structs in `Pool.sol` contract 
```
    struct LendingDetails {
        *uint64 marginCallEndTime;*
        uint256 effectiveInterestWithdrawn;
        uint256 extraLiquidityShares;
    }

    // Pool constants
    struct PoolConstants {
        *uint64 loanStartTime;*
        *uint128 borrowRate;*
        address borrower;
        address collateralAsset;
        address borrowAsset;
        address poolSavingsStrategy; // invest contract
        address lenderVerifier;
        uint256 loanWithdrawalDeadline;
        uint256 idealCollateralRatio;
        uint256 noOfRepaymentIntervals;
        uint256 repaymentInterval;
        uint256 borrowAmountRequested;
    }
```
-  Restructured the structs in `Repayments.sol` contract 
```
    struct RepaymentVariables {
        *bool isLoanExtensionActive;*
        uint256 repaidAmount;
        uint256 loanDurationCovered;
        uint256 loanExtensionPeriod; // period for which the extension was granted, ie, if loanExtensionPeriod is 7 * 10**30, 7th instalment can be repaid by 8th instalment deadline
    }

    struct RepaymentConstants {
        *uint128 gracePenaltyRate;*
        *uint128 gracePeriodFraction;*
        *uint128 borrowRate;*
        address repayAsset;
        uint256 numberOfTotalRepayments; // using it to check if RepaymentDetails Exists as repayment Interval!=0 in any case
        uint256 loanDuration;
        uint256 repaymentInterval;
        uint256 loanStartTime;
    }
```
- Restructured the structs in `Verification.sol`
```
struct LinkedAddress {
        *uint64 activatesAt;*
        address masterAddress;
    }
[...]

PR #181, opened 2021-12-27
lime-77, Issue 40: Inlining logic used once

KalpitaMandal:
# Description
Stacked pull request on #180
Reduce code size and gas expenditure by inlining internal functions that are only used once throughout your contract.

In some of your contracts, like AaveYield.sol, you have an initialize function that calls the internal functions updateSavingsAccount and updateAaveAddresses. These two functions are only called in the initialize function, so you can save some gas and code size by simply inlining their logic inside initialize. The tradeoff, of course, is that you lose readability, and because initialize will only be called once, perhaps it's not worth the tradeoff.

# Integrations Checklist

# Changelog

- Adding inline functions for initialize functions instead of internal function calls(Issue 40: Inlining logic used once
Ref: https://github.com/code-423n4/2021-12-sublime-findings/issues/40)

# Event Signature Changes

# Function Signature Changes
- Following changes have been done for all the yield contracts (Aave, Compound, Yearn and No-yield)
  - `function updateSavingsAccount(address payable _savingsAccount) external onlyOwner` (Removed)
  - `function _updateSavingsAccount(address payable _savingsAccount) internal` (Removed)
- `function updateVerification(address _verification) external onlyOwner` (Removed)
- `function _updateVerification(address _verification) internal` (Removed)

# Features

# Events


prateekreddy:
It is probably not necessary in view of readability

PR #188, opened 2021-12-28
Lime-132, Issue 168: Ether can be locked in the PoolFactory

KalpitaMandal:
# Description
Stacked pull request on #186
If a borrower calls the `createPool` function with a non-zero value, but also includes an ERC20 token address for `_collateralToken`, then the Ether value sent will be locked in the `PoolFactory` contract forever.
- [createPool L260-317](https://github.com/code-423n4/2021-12-sublime/blob/main/contracts/Pool/PoolFactory.sol#L260-L317)

In the `_createPool` function, a `_collateralToken` address other than the zero address will set the `amount` variable to zero. That `amount` variable will be passed to `create2` which will send 0 wei to the newly created `Pool` contract.
```
// _createPool L349
uint256 amount = _collateralToken == address(0) ? _collateralAmount : 0;
```
# Integrations Checklist

# Changelog

- Adding a require statement to ensure no ether is sent when ETH is not set as collateral Token(Issue 168: Ether can be locked in the PoolFactory
Ref: https://github.com/code-423n4/2021-12-sublime-findings/issues/168)

# Event Signature Changes

# Function Signature Changes

# Features

# Events


prateekreddy (approved):
lgtm

PR #216, opened 2022-01-05
Fixes-lime-68, Wrong emergencyWithdraw logic

KalpitaMandal:
# Description
Stacked pull request on #199
The method, `emergencyWithdraw` inside the contract `yield/NoYield` doesn't work as expected, the `transfer` was done with received value, and it should be done with amount, so the emergencyWithdraw never will work, always it will send 0 tokens.

# Integrations Checklist

# Changelog

- Returning correct return values for emergency withdraw(Wrong emergencyWithdraw logic
There is a bug in emergency withdraw logic of no yield function as specified in issue https://github.com/code-423n4/2021-12-sublime-findings/issues/4)

# Event Signature Changes

# Function Signature Changes

# Features

# Events


contracts/yield/NoYield.sol
```diff
@@ -77,9 +77,8 @@ contract NoYield is IYield, Initializable, OwnableUpgradeable, ReentrancyGuard {
      */
     function emergencyWithdraw(address _asset, address payable _wallet) external onlyOwner returns (uint256 received) {
         require(_wallet != address(0), 'cant burn');
-        uint256 amount = IERC20(_asset).balanceOf(address(this));
+        received = IERC20(_asset).balanceOf(address(this));
```
prateekreddy, line 80: remove received var from the returns declaration and return received at the end of function

prateekreddy (changes requested):

contracts/yield/NoYield.sol
KalpitaMandal: Done

prateekreddy (approved):

PR #249, opened 2022-01-11
CS Audit, Lime 220, Removed unused enum state

saxenism:
## Description

In CreditLine the CreditLineStatus has the definition CANCELLED. This definition is never used and therefore can be removed.

## Integration Checklist

- [ ] Make sure all existing tests should pass
- [ ] Gas costs reduction can be compared using gasReport.md

## Changelog

An unused enum state was removed from CreditLines.

### DIFF
--- a/contracts/CreditLine/CreditLine.sol
+++ b/contracts/CreditLine/CreditLine.sol
@@ -26,7 +26,6 @@ contract CreditLine is ReentrancyGuard, OwnableUpgradeable {
         REQUESTED,
         ACTIVE,
         CLOSED,
-        CANCELLED,
         LIQUIDATED
     }
 


contracts/CreditLine/CreditLine.sol
```diff
@@ -26,7 +26,6 @@ contract CreditLine is ReentrancyGuard, OwnableUpgradeable {
         REQUESTED,
         ACTIVE,
         CLOSED,
-        CANCELLED,
```
prateekreddy, line 29: Can you add a function to cancel the credit line when it is not yet accepted ?

prateekreddy (changes requested):

contracts/CreditLine/CreditLine.sol
saxenism: Please refer to https://github.com/sublime-finance/sublime-v1/pull/257

Let me know if you want me to make that changes here.

prateekreddy:
duplicate of #257

PR #250, opened 2022-01-11
CS Audit, Lime 215, Removed Superfluous check

saxenism:
## Description

in Verfication.unregisterMasterAddress the _verifier argument is obsolete including the associated check and use. All could be re-written with msg.sender as the require enforces msg.sender == _verifier.

## Integration Checklist

- [ ] All existing tests should still pass
- [ ] Savings in gas cost can be assessed using gasReport.md

## Change Log

A superfluous check was removed from the Verification contract.


contracts/Verification/Verification.sol
```diff
@@ -79,12 +79,12 @@ contract Verification is Initializable, IVerification, OwnableUpgradeable {
     /// @dev unregistering master address doesn't affect linked addreses mapping to master address, though they would not be verified by this verifier anymore
     /// @param _masterAddress address which is being unregistered
     /// @param _verifier verifier address from which master address is unregistered
-    function unregisterMasterAddress(address _masterAddress, address _verifier) external override {
+    function unregisterMasterAddress(address _masterAddress) external override {
         if (msg.sender != super.owner()) {
-            require(masterAddresses[_masterAddress][msg.sender] || msg.sender == _verifier, 'V:UMA-Invalid verifier');
+            require(masterAddresses[_masterAddress][msg.sender] != 0, 'V:UMA-Invalid verifier');
```
prateekreddy, line 84: it should be != address(0) as the types won't match

contracts/Verification/Verification.sol
```diff
@@ -79,12 +79,12 @@ contract Verification is Initializable, IVerification, OwnableUpgradeable {
     /// @dev unregistering master address doesn't affect linked addreses mapping to master address, though they would not be verified by this verifier anymore
     /// @param _masterAddress address which is being unregistered
     /// @param _verifier verifier address from which master address is unregistered
-    function unregisterMasterAddress(address _masterAddress, address _verifier) external override {
+    function unregisterMasterAddress(address _masterAddress) external override {
```
prateekreddy, line 82: verification contract admin also can remove verifier, so in that case, _verifier of the master address should be passed as arg

contracts/Verification/Verification.sol
```diff
@@ -79,12 +79,12 @@ contract Verification is Initializable, IVerification, OwnableUpgradeable {
...
         if (msg.sender != super.owner()) {
-            require(masterAddresses[_masterAddress][msg.sender] || msg.sender == _verifier, 'V:UMA-Invalid verifier');
+            require(masterAddresses[_masterAddress][msg.sender] != 0, 'V:UMA-Invalid verifier');
         }
-        delete masterAddresses[_masterAddress][_verifier];
-        emit UserUnregistered(_masterAddress, _verifier, msg.sender);
+        delete masterAddresses[_masterAddress][msg.sender];
```
prateekreddy, line 86: This and the line below should use _verifier to cover the case when verification admin unregisters the master address

prateekreddy (changes requested):

contracts/Verification/Verification.sol
saxenism: Hmm, I think there is some misunderstanding here.
`masterAddresses` is defined as follows:
` mapping(address => mapping(address => uint256)) public masterAddresses`
So, this returns the time (in uint256) since the verification has been active and not any addresses. So, I think, this check is fine.

Please confirm.

saxenism: I've changed the definition of the `masterAddresses` mapping.  `!=0` is consistent with that definition. 

Please check that out.

saxenism:
> ## Description
> in Verfication.unregisterMasterAddress the _verifier argument is obsolete including the associated check and use. All could be re-written with msg.sender as the require enforces msg.sender == _verifier.
> 
> ## Integration Checklist
> * [ ]  All existing tests should still pass
> * [ ]  Savings in gas cost can be assessed using gasReport.md
> 
> ## Change Log
> A superfluous check was removed from the Verification contract.

Actually I went through the Lime issue (LIME-215) once again and I don't think it is valid. True that we don't need the _verifier parameter to check if the function called is the verifier or not, we could have done `verifiers[msg.sender` . But the thing is that, both the `Verification` admin and the particular `verifier` can call this function. So, at the time of deleting the mapping, we need to pass the address of the verifier. And if we are not sure that the function was called by the verifier or the admin, how can we remove the verifier parameter.

contracts/Verification/Verification.sol
saxenism: True.  I don't think the Lime issue that this PR refers to (Lime-215) is valid. Please see my latest comment below.

contracts/Verification/Verification.sol
saxenism: True.  I don't think the Lime issue that this PR refers to (Lime-215) is valid. Please see my latest comment below.

saxenism:
> ## Description
> in Verfication.unregisterMasterAddress the _verifier argument is obsolete including the associated check and use. All could be re-written with msg.sender as the require enforces msg.sender == _verifier.
> 
> ## Integration Checklist
> * [ ]  All existing tests should still pass
> * [ ]  Savings in gas cost can be assessed using gasReport.md
> 
> ## Change Log
> A superfluous check was removed from the Verification contract.

PR began with this code change only, but later, after discussions with the team we decided to include as many changes from the `Verification.sol` contract of `master` to `v1-alpha`. So, in this I've changed the definition of the `masterAddresses` mapping and introduced the concept of `activationDelay` and then propagated those changes in the tests.

contracts/Verification/Verification.sol
saxenism: This is simply because the mapping `masterAddresses` is defined as follows:

```
    /// @notice Maps masterAddress with the verifier that was used to verify it and the time when master address is active
    /// @dev Mapping is from masterAddress -> verifier -> activationTime
    /// @return Verifier used to verify the given master address
    mapping(address => mapping(address => uint256)) public masterAddresses;
```

So, if this mapping gives any result other than zero, for `masterAddresses[_masterAddress][msg.sender]` then that already implies that `msg.sender` is the `verifier` for `_masterAddress`. So the next check of `msg.sender == _verifier` was actually redundant.

But let me know if you think we should still keep it, I'll add that again then.

prateekreddy (changes requested):

contracts/Verification/Verification.sol
```diff
@@ -142,6 +142,7 @@ contract Verification is Initializable, IVerification, OwnableUpgradeable {
         require(pendingLinkAddresses[msg.sender][_masterAddress], 'V:LA-No pending request');
         _linkAddress(msg.sender, _masterAddress);
```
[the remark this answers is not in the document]
saxenism: Correct. I've made the relevant change. Please review again.

prateekreddy (approved):

PR #251, opened 2022-01-11
CS Audit, Lime 191, Renamed poolFactory modifier

saxenism:
## Description

The PoolFactory contract implements a modifier as shown below: 

modifier onlyBorrower(address _verifier) { require( IVerification(userRegistry).isUser(msg.sender, _verifier), 'PoolFactory::onlyBorrower - Only a valid Borrower can create Pool' ); _; } 

The name onlyBorrower implies that only a borrower of a pool can call functions protected by this modifier, however, the modifier only checks that the msg.sender has been verified by _verifier and not if it is a borrower of a pool.

## Integration Checklist

- [ ] The exisiting tests should not break.

## Change Log

The name of the `onlyBorrower` modifier in PoolFactory.sol has been changed to a more appropriate `onlyVerified`.

### DIFF
--- a/contracts/Pool/PoolFactory.sol
+++ b/contracts/Pool/PoolFactory.sol
@@ -153,12 +153,12 @@ contract PoolFactory is Initializable, OwnableUpgradeable, IPoolFactory {
     }
 
     /**
-     * @notice functions affected by this modifier can only be invoked by the borrow of the Pool
+     * @notice functions affected by this modifier can only be invoked by the borrower (verified user) of the Pool
      */
-    modifier onlyBorrower(address _verifier) {
+    modifier onlyVerified(address _verifier) {
         require(
             IVerification(userRegistry).isUser(msg.sender, _verifier),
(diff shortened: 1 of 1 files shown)


contracts/Pool/PoolFactory.sol
```diff
@@ -153,12 +153,12 @@ contract PoolFactory is Initializable, OwnableUpgradeable, IPoolFactory {
     }
 
     /**
-     * @notice functions affected by this modifier can only be invoked by the borrow of the Pool
+     * @notice functions affected by this modifier can only be invoked by the borrower (verified user) of the Pool
```
prateekreddy, line 156: actually let's probably remove borrower stuff, we just need to say that it is invoked by verified user na... borrower stuff is secondary and might not apply everywhere.

prateekreddy (changes requested):

contracts/Pool/PoolFactory.sol
saxenism: True. Made the changes. Please review again.

prateekreddy (approved):

PR #252, opened 2022-01-11
CS Audit, Lime 192, Added Sanity checks on share transfer

saxenism:
## Description

The function transfer does not perform sanity checks for the parameter _to. Hence, it is possible to send to address zero and effectively "burn" the shares from the saving strategies during transfer.

## Integration Checklist

- [ ] Add tests to check the sanity checks added to all the different transfer and withdraw functions
- [ ] Add tests to check the functioning of the `burn` function

## Change Log

1. Sanity check were added to the following functions in `SavingsAccount.sol` contract : 
    1. `SavingsAccount.withdraw`
    2. `SavingsAccount.withdrawAll`
    3. `SavingsAccount.transfer`
    4. `SavingsAccount.transferFrom`
    5. `SavingsAccount._transfer`

2. A new `burn` function was added in the `SavingsAccount.sol` contract. This function is used to burn a user's token shares from their savings account.

3. The `burn` function was also added in the `ISavingsAccount` interface.


contracts/SavingsAccount/SavingsAccount.sol
```diff
@@ -270,6 +270,7 @@ contract SavingsAccount is ISavingsAccount, Initializable, OwnableUpgradeable, R
         address _token,
         address payable _to
     ) internal {
+        require(_to != address(0), 'The to address should be a valid address');
```
prateekreddy, line 273: In this case, we need a seperate function for burn... Can you look into it ?

prateekreddy (changes requested):

contracts/SavingsAccount/SavingsAccount.sol
saxenism: Yes, agreed. Added the burn function. This burn function only takes care of burning the corresponding shares of the user's tokens in the savings account particular strategy. Burning of directly the tokens is not implemented and left to the users.

contracts/SavingsAccount/SavingsAccount.sol
```diff
@@ -449,6 +458,36 @@ contract SavingsAccount is ISavingsAccount, Initializable, OwnableUpgradeable, R
...
+    */
+    function burn(
+        uint256 _amount, 
+        address _token, 
+        address _strategy
+    ) external override returns(uint256) {
+        require(_amount != 0, "SavingsAccount::burn _amount cannot be zero");
+        require(IStrategyRegistry(strategyRegistry).registry(_strategy), 'SavingsAccount::burn strategy do not exist');
+
+        if(_strategy != address(0)) {
```
ritik99, line 477: We're using no yield strat now, is this still valid?

contracts/SavingsAccount/SavingsAccount.sol
```diff
@@ -449,6 +458,36 @@ contract SavingsAccount is ISavingsAccount, Initializable, OwnableUpgradeable, R
...
+            _amount = IYield(_strategy).getSharesForTokens(_amount, _token);
+        }
+
+        balanceInShares[msg.sender][_token][_strategy] = balanceInShares[msg.sender][_token][_strategy].sub(
```
ritik99, line 481: check the current balance first before subtraction

saxenism: Yes, if you see the full statement about subtraction here: https://github.com/sublime-finance/sublime-v1/blob/chainsec-lime-192/contracts/SavingsAccount/SavingsAccount.sol#L483
Then you'll see that I've added a check to see that the subtraction only goes through if there is enough balance.

Let me know if you think this is not appropriate, I'll replace this with a `require` statement in that case.

contracts/SavingsAccount/SavingsAccount.sol
saxenism: True, there is an `NoYield` contract now and we could have called that contract like `IYield(_strategy).getSharesForTokens(_amount, _token)` but as you can see in the `NoYield` contract itself here https://github.com/sublime-finance/sublime-v1/blob/chainsec-lime-192/contracts/yield/NoYield.sol#L176 that it will only waste gas, as no operation is being performed in the `getSharesForTokens` function in `NoYield` hence I think this change is pretty much valid.

saxenism: Hmm, I get what you're saying now. The `if` block was redundant. Have removed that.

contracts/SavingsAccount/SavingsAccount.sol
```diff
@@ -272,6 +275,7 @@ contract SavingsAccount is ISavingsAccount, Initializable, OwnableUpgradeable, R
         address _token,
         address payable _to
     ) internal {
+        require(_to != address(0), "SavingsAccounts::_transfer _to address should be non-zero");
```
prateekreddy, line 278: We are doing multiple check on to!= address(0), for example, withdraw function checks for this twice. Let's optimize that part

contracts/SavingsAccount/SavingsAccount.sol
```diff
@@ -449,6 +458,34 @@ contract SavingsAccount is ISavingsAccount, Initializable, OwnableUpgradeable, R
...
+    @param _strategy The address of the strategy from which the _token's shares will be burnt
+    @return _amount The amount of _token shares that were burnt
+    */
+    function burn(
+        uint256 _amount, 
+        address _token, 
+        address _strategy
+    ) external override returns(uint256) {
+        require(_amount != 0, "SavingsAccount::burn _amount cannot be zero");
+        require(IStrategyRegistry(strategyRegistry).registry(_strategy), 'SavingsAccount::burn strategy do not exist');
```
prateekreddy, line 475: Any specific reason we are checking for strategy here ?

prateekreddy (changes requested):

contracts/SavingsAccount/SavingsAccount.sol
saxenism: The only reason is for sanitising the strategy address, so that no-one passes an invalid strategy address to burn their shares from.

Should I remove this check?

contracts/SavingsAccount/SavingsAccount.sol
saxenism: Ha true, the `function withdraw` does check if the `_to` address is not zero. But I re-checked that no function which is calling this specific `function _transfer` function is making this `_to` address check.

Please review again.

contracts/SavingsAccount/SavingsAccount.sol
prateekreddy: yeah, because it is possible that user wants to burn tokens if the strategy is considered invalid. So probably this is not needed.

saxenism: Cool. Removed this check.

prateekreddy (approved):

PR #253, opened 2022-01-11
CS Audit, Lime 181, Stacked requires together

saxenism:
## Description

The following function in SavingsAccount 

deposit(uint256 _amount, address _token, address _strategy, address _to) 

does a single sanity check for the _to parameter, then calls the internal function _deposit which does another sanity check for the amount parameter. Next, function _depositToYield is called which does another sanity check for the initial parameter _strategy. 

It is a good practice to do all such checks first in the initial function when possible, e.g., deposit, instead of splitting them among multiple functions. Other examples: 

• require(_lender != _borrower) in _createRequest. 

• require(creditLineConstants[_id].lender != msg.sender) in _depositCollateral.

## Integration Checklist

- [ ] The existing tests shouldn't break

## Change Log

Different require statements were stacked together in the SavingsAccount.sol smart contract.

### DIFF
--- a/contracts/SavingsAccount/SavingsAccount.sol
+++ b/contracts/SavingsAccount/SavingsAccount.sol
@@ -158,6 +158,7 @@ contract SavingsAccount is ISavingsAccount, Initializable, OwnableUpgradeable, R
     ) external override nonReentrant {
         require(_currentStrategy != _newStrategy, 'SavingsAccount::switchStrategy Same strategy');
         require(_amount != 0, 'SavingsAccount::switchStrategy Amount must be greater than zero');
+        require(IStrategyRegistry(strategyRegistry).registry(_newStrategy), 'SavingsAccount::deposit strategy do not exist');
 
         _amount = IYield(_currentStrategy).getSharesForTokens(_amount, _token);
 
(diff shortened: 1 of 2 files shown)


contracts/SavingsAccount/SavingsAccount.sol
```diff
@@ -132,7 +134,6 @@ contract SavingsAccount is ISavingsAccount, Initializable, OwnableUpgradeable, R
         address _token,
         address _strategy
     ) internal returns (uint256 _sharesReceived) {
-        require(IStrategyRegistry(strategyRegistry).registry(_strategy), 'SavingsAccount::deposit strategy do not exist');
```
prateekreddy, line 135: `_depositToYield` has been used in multiple functions. So the check needs to be migrated to other functions as well

prateekreddy (changes requested):

contracts/SavingsAccount/SavingsAccount.sol
saxenism: Yes, correct. Migrated the require function to the required functions. Please review once again.

prateekreddy (approved):

PR #257, opened 2022-01-11
CS Audit, Lime 206, Gas opt in creditline closure

saxenism:
## Description

A credit line uses two struct variables CreditLineConstants and CreditLineVariables which occupy storage when a credit line is created. When a credit line is closed, the status variable is set to CLOSED but the occupied storage is not freed.

## Integration Checklist

- [ ] Write tests calling public/external functions after a particular creditline has been closed and make sure that no function throws unexpected values
- [ ] Savings in gas cost can be compared using gasReport.md

## Change Log
Once a credit line has been closed, the associated creditLineConstants and creditLineVariables are deleted.

### DIFF
--- a/contracts/CreditLine/CreditLine.sol
+++ b/contracts/CreditLine/CreditLine.sol
@@ -872,6 +872,8 @@ contract CreditLine is ReentrancyGuard, OwnableUpgradeable {
         require(creditLineVariables[_id].principal == 0, 'CreditLine: Cannot be closed since not repaid.');
         require(creditLineVariables[_id].interestAccruedTillLastPrincipalUpdate == 0, 'CreditLine: Cannot be closed since not repaid.');
         creditLineVariables[_id].status = CreditLineStatus.CLOSED;
+        delete creditLineConstants[_id];
+        delete creditLineVariables[_id];
         emit CreditLineClosed(_id);
     }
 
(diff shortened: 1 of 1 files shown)


saxenism:
Topic for discussion @ritik99 @prateekreddy @prateekreddy @KalpitaMandal 

With the above PR, if approved, I don't think we would really need the `CLOSED` enum state too. Let me know what you all think.

contracts/CreditLine/CreditLine.sol
```diff
@@ -872,6 +872,8 @@ contract CreditLine is ReentrancyGuard, OwnableUpgradeable {
         require(creditLineVariables[_id].principal == 0, 'CreditLine: Cannot be closed since not repaid.');
         require(creditLineVariables[_id].interestAccruedTillLastPrincipalUpdate == 0, 'CreditLine: Cannot be closed since not repaid.');
         creditLineVariables[_id].status = CreditLineStatus.CLOSED;
```
prateekreddy, line 874: This line needs to be removed

prateekreddy (changes requested):

contracts/CreditLine/CreditLine.sol
saxenism: Removed. Please review again.

saxenism:
> Topic for discussion @ritik99 @prateekreddy
> 
> With the above PR, if approved, I don't think we would really need the `CLOSED` enum state in the `creditLineStatus` too. Let me know what you all think.

The above topic was discussed on team meet and I have removed the `CLOSED` state as well as introduced a `cancel` function to cancel creditline requests without adding a `CANCELLED` state.

PR #258, opened 2022-01-11
CS Audit, Lime 170, Added description of the return value

saxenism:
## Description

Return value of `SavingsAccount.transferFrom` lacks description

## Change Log

Natpsec comment of `SavingsAccount.transferFrom` was improved.

### DIFF
--- a/contracts/SavingsAccount/SavingsAccount.sol
+++ b/contracts/SavingsAccount/SavingsAccount.sol
@@ -409,6 +409,8 @@ contract SavingsAccount is ISavingsAccount, Initializable, OwnableUpgradeable, R
      * @param _strategy address of the strategy from which tokens are transferred
      * @param _from address from whose allowance tokens are transferred
      * @param _to address of the user tokens are transferred to
+     * @return the amount of tokens in terms of _token that will be transferred from the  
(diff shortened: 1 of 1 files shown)


contracts/SavingsAccount/SavingsAccount.sol
```diff
@@ -409,6 +409,8 @@ contract SavingsAccount is ISavingsAccount, Initializable, OwnableUpgradeable, R
      * @param _strategy address of the strategy from which tokens are transferred
      * @param _from address from whose allowance tokens are transferred
      * @param _to address of the user tokens are transferred to
+     * @return the amount of tokens in terms of _token that will be transferred from the
```
prateekreddy, line 412: the amount is not in _token but the LP token for _token in strategy

prateekreddy (changes requested):

contracts/SavingsAccount/SavingsAccount.sol
saxenism: Yes, correct. Had gotten a bit confused. Please review again.

prateekreddy (approved):

PR #259, opened 2022-01-11
CS Audit, Lime 173, Corrected the docs

saxenism:
## Description

The function Repayments.getInterestOverdue has a wrong description since it considers a case when there is a loan extension.

## Change Log

Erroneous function description was corrected.

### DIFF
--- a/contracts/Pool/Extension.sol
+++ b/contracts/Pool/Extension.sol
@@ -2,7 +2,7 @@
 pragma solidity 0.7.0;
 
 import '@openzeppelin/contracts-upgradeable/proxy/Initializable.sol';
-import '@openzeppelin/contracts/token/ERC20/SafeERC20.sol';
+import '@openzeppelin/contracts/math/SafeMath.sol';
 import '../interfaces/IPool.sol';
 import '../interfaces/IPoolFactory.sol';
 import '../interfaces/IExtension.sol';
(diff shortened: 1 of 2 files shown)


contracts/Pool/Extension.sol
```diff
@@ -2,7 +2,7 @@
 pragma solidity 0.7.0;
 
 import '@openzeppelin/contracts-upgradeable/proxy/Initializable.sol';
-import '@openzeppelin/contracts/token/ERC20/SafeERC20.sol';
+import '@openzeppelin/contracts/math/SafeMath.sol';
```
prateekreddy, line 5: Why was this change necessary ?

prateekreddy (changes requested):

contracts/Pool/Extension.sol
saxenism: According to an earlier issue, we found that `Extension.sol` imports SafeERC20.sol, but it is not used in the contract anywhere, so I had first removed it, but then the contract was using SafeMath for uint256.... so, I had to import `math/SafeMath`.
This essentially means that before this, the only use of `token/ERC20/SafeERC20` was enabling the use of SafeMath and nothing else.

Apart from that, since you've already approved PR#248 (https://github.com/sublime-finance/sublime-v1/pull/248) not making this change would have created merge conflict also.

prateekreddy (approved):

PR #260, opened 2022-01-11
CS Audit, Lime 174, Improved documentation

saxenism:
## Description

In AaveYield.sol
• The specification for the `unlockShares` refers to the asset as the underlying token, but the code assumes to be the Aave token. 

• In _withdrawETH the comment lock collateral is incorrect.

## Change Log

Erroneous documentation were corrected.

## Function Signature Change

The function signature for `function unlockShares` was changed from 

```
function unlockShares(address asset, uint256 amount) external returns (uint256 received);
```

to

```
function unlockShares(address asset, uint256 amount) external returns (uint256 received, address strategyTokens);
```


contracts/yield/AaveYield.sol
```diff
@@ -231,7 +231,7 @@ contract AaveYield is IYield, Initializable, OwnableUpgradeable, ReentrancyGuard
 
     /**
      * @notice Used to unlock shares
-     * @param asset the address of underlying token
+     * @param asset the address of underlying token, which in this would always be Aave tokens
```
prateekreddy, line 234: Let's instead use underlying asset only here and change all the instances where this is used

prateekreddy: It will save gas costs as we are making multiple calls to yield which can be avoided

prateekreddy (changes requested):

contracts/yield/AaveYield.sol
saxenism: This is done. Please check again.

contracts/yield/AaveYield.sol
```diff
@@ -245,20 +245,23 @@ contract AaveYield is IYield, Initializable, OwnableUpgradeable, ReentrancyGuard
...
+     * @param asset the address of token locked
      * @param amount the amount of shares to unlock
      * @return received amount of shares received
+     * @return address of Aave shares of asset 
      **/
-    function unlockShares(address asset, uint256 amount) external override onlySavingsAccount nonReentrant returns (uint256) {
+    function unlockShares(address asset, uint256 amount) external override onlySavingsAccount nonReentrant returns (uint256, address) {
         if (amount == 0) {
-            return 0;
+            return (0, address(0));
```
prateekreddy, line 255: Let's still return the correct liquidityToken for correctness

prateekreddy (changes requested):

contracts/yield/AaveYield.sol
saxenism: Yes, correct. Changed the return value. 

Please review again.

prateekreddy (approved):

PR #265, opened 2022-01-11
CS lime-189, Inefficient Struct Packing

KalpitaMandal:
# Description

The struct `ExtensionVariables` has three variables of type `uint256`. At least the time and interval variables could be packed together to save a slot as it is highly unlikely, that these variables need a full word.

# Integrations Checklist

# Changelog

- Efficient struct packing in `Extension.sol`.

# Event Signature Changes

# Function Signature Changes

# Features

# Events

### DIFF
--- a/contracts/Pool/Extension.sol
+++ b/contracts/Pool/Extension.sol
@@ -18,10 +18,10 @@ contract Extension is Initializable, IExtension {
 
     struct ExtensionVariables {
         bool hasExtensionPassed;
+        uint64 extensionVoteEndTime;
+        mapping(address => uint64) lastVotedExtension;
         uint256 totalExtensionSupport;
-        uint256 extensionVoteEndTime;
         uint256 repaymentInterval;
-        mapping(address => uint256) lastVotedExtension;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


contracts/Pool/Extension.sol
```diff
@@ -18,10 +18,10 @@ contract Extension is Initializable, IExtension {
 
     struct ExtensionVariables {
         bool hasExtensionPassed;
+        uint64 extensionVoteEndTime;
+        mapping(address => uint64) lastVotedExtension;
         uint256 totalExtensionSupport;
-        uint256 extensionVoteEndTime;
         uint256 repaymentInterval;
```
prateekreddy, line 24: May be we can combine repaymentInterval as well with the first 2 vars, because bool takes 8 bits only.

prateekreddy (changes requested):

contracts/Pool/Extension.sol
```diff
@@ -18,10 +18,10 @@ contract Extension is Initializable, IExtension {
 
     struct ExtensionVariables {
         bool hasExtensionPassed;
+        uint64 extensionVoteEndTime;
+        mapping(address => uint64) lastVotedExtension;
```
prateekreddy, line 22: mapping can be uint256 only, we will end up spending more gas if we use uint64

KalpitaMandal: reverted to uint256

contracts/Pool/Extension.sol
KalpitaMandal: Packing `repaymentInterval` further using uint128

prateekreddy (approved):

PR #266, opened 2022-01-11
CS lime-190, Inefficient Struct PriceData

KalpitaMandal:
# Description
The struct PriceData has two variables and is defined as: 

struct PriceData { address oracle; uint256 decimals; } 

decimals stores the number of decimals that the price is returned, hence its range is limited (currently the code returns prices with 30 decimals). Given that address occupies 20 bytes, the struct can be packed together in a single storage slot.

# Integrations Checklist

# Changelog

- Efficient struct packing in `PriceOracle.sol`.

# Event Signature Changes

# Function Signature Changes

# Features

# Events

### DIFF
--- a/contracts/PriceOracle.sol
+++ b/contracts/PriceOracle.sol
@@ -14,8 +14,8 @@ contract PriceOracle is Initializable, OwnableUpgradeable, IPriceOracle {
 
     uint32 uniswapPriceAveragingPeriod;
     struct PriceData {
+        uint64 decimals;
         address oracle;
-        uint256 decimals;
     }
     /**
      * @notice stores the price oracle and its decimals for chainlink feeds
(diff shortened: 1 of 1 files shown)


contracts/PriceOracle.sol
```diff
@@ -14,8 +14,8 @@ contract PriceOracle is Initializable, OwnableUpgradeable, IPriceOracle {
 
     uint32 uniswapPriceAveragingPeriod;
     struct PriceData {
+        uint64 decimals;
```
prateekreddy, line 17: order can still be the same right, any specific reason for changing the order of the struct

KalpitaMandal: Order does not matter here but it is standard that the smaller data types are defined first

PR #267, opened 2022-01-11
CS lime-193, No Event Emitted on Emergency Withdrawal

KalpitaMandal:
# Description
All saving strategies implement a function emergencyWithdraw(address _asset, address payable _wallet) which allows the owner to transfer all tokens of an _asset to an arbitrary address _wallet. Due to the sensitivity of this operation, an event should be emitted when the emergency withdraw executes.

# Integrations Checklist

# Changelog

- Adding EmergencyWithdraw event in `IYield.sol`, `AaveYield.sol`, `CompoundYield.sol`, `YearnYield.sol` and `NoYield.sol`.
- Adding comments for `emergencyWithdraw` function in all strategy contracts.

# Event Signature Changes

# Function Signature Changes

# Features

# Events

### DIFF
--- a/contracts/interfaces/IYield.sol
+++ b/contracts/interfaces/IYield.sol
@@ -2,6 +2,15 @@
 pragma solidity 0.7.0;
 
 interface IYield {
+
+    /**
+     * @notice emitted when all tokens are withdrawn, in case of emergencies
+     * @param asset address of the token being withdrawn
+     * @param withdrawTo address of the wallet to which tokens are withdrawn
+     * @param tokensReceived amount of tokens received
+     */
+    event EmergencyWithdraw(address indexed asset, address indexed withdrawTo, uint256 tokensReceived);
+    
     /**
      * @dev emitted when tokens are locked
      * @param user the address of user, tokens locked for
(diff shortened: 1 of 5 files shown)


contracts/interfaces/IYield.sol
```diff
@@ -2,6 +2,15 @@
 pragma solidity 0.7.0;
 
 interface IYield {
+
+    /**
+     * @notice emitted when all tokens are withdrawn, in case of emergencies
+     * @param asset address of the token being withdrawn
+     * @param withdrawTo address of the wallet to which tokens are withdrawn
+     * @param tokensReceived amount of tokens received
+     */
```
prateekreddy, line 11: Let's not add emergency withdraw to IYield because we do not want to enforce this function in all the yield contracts. Though we are doing here, but eventually we won't.

prateekreddy (changes requested):

contracts/interfaces/IYield.sol
KalpitaMandal: Sure, adding it to the individual contracts. We can remove them later when not required.

prateekreddy (approved):

PR #268, opened 2022-01-11
CS lime-205, Possible Gas Optimization in Mappings

KalpitaMandal:
# Description
Multiple contracts of the system use mappings in the format: mapping(key_type => bool). Solidity uses a word (256 bits) for each stored value and performs some additional operations when operating bool values (due to masking). Therefore, using uint instead of bool is slightly more efficient.

# Integrations Checklist

# Changelog

- Using mappings to uint instead of bool to avoid excess gas cost.

# Event Signature Changes

# Function Signature Changes
- `function registry(address _strategy) external view returns (*uint*)` (IStrategyRegistry.sol)
- `function poolRegistry(address pool) external view returns (*uint*)` (IPoolFactory.sol)
# Features

# Events


contracts/Pool/PoolFactory.sol
```diff
@@ -107,17 +107,17 @@ contract PoolFactory is Initializable, OwnableUpgradeable, IPoolFactory {
     /*
      * @notice Used to mark assets supported for borrowing
      */
-    mapping(address => bool) isBorrowToken;
+    mapping(address => uint) isBorrowToken;
```
prateekreddy, line 110: use uint256 instead of uint for consistency everywhere

contracts/SavingsAccount/SavingsAccount.sol
```diff
@@ -156,6 +156,7 @@ contract SavingsAccount is ISavingsAccount, Initializable, OwnableUpgradeable, R
         address _newStrategy
     ) external override nonReentrant {
         require(_currentStrategy != _newStrategy, 'SavingsAccount::switchStrategy Same strategy');
+        require(IStrategyRegistry(strategyRegistry).registry(_newStrategy) != 0, 'SavingsAccount::_newStrategy do not exist');
```
prateekreddy, line 159: This check didn't exist before ?

prateekreddy (changes requested):

contracts/SavingsAccount/SavingsAccount.sol
KalpitaMandal: Not in v1-alpha

prateekreddy (approved):

PR #269, opened 2022-01-11
CS lime-209, Recalculation of Shares

KalpitaMandal:
# Description
The _depositCollateralFromSavingsAccount in the CreditLine contract calculates the shares transferred and then adds them into the balance:

collateralShareInStrategy[_id][_strategy] = collateralShareInStrategy[_id][_strategy].add( _liquidityShares.mul(_tokensToTransfer).div(_tokenInStrategy) );

Instead of recalculating the shares, the function can use the return value of the external call transferFrom which returns the share amount.

# Integrations Checklist

# Changelog

- Using return values instead of re-calculating the shares in `_depositCollateralFromSavingsAccount` in the `CreditLine.sol`.

# Event Signature Changes

# Function Signature Changes

# Features

# Events


contracts/CreditLine/CreditLine.sol
```diff
@@ -488,18 +488,16 @@ contract CreditLine is ReentrancyGuard, OwnableUpgradeable {
...
             uint256 _tokensToTransfer = _tokenInStrategy;
             if (_activeAmount.add(_tokenInStrategy) >= _amount) {
                 _tokensToTransfer = (_amount.sub(_activeAmount));
             }
             _activeAmount = _activeAmount.add(_tokensToTransfer);
-            _savingsAccount.transferFrom(_tokensToTransfer, _collateralAsset, _strategy, _sender, address(this));
 
             collateralShareInStrategy[_id][_strategy] = collateralShareInStrategy[_id][_strategy].add(
-                _liquidityShares.mul(_tokensToTransfer).div(_tokenInStrategy)
+                _savingsAccount.transferFrom(_tokensToTransfer, _collateralAsset, _strategy, _sender, address(this))
```
prateekreddy, line 500: Create a local variable to hold the return value of transferFrom so that code is more readable

prateekreddy (changes requested):

contracts/CreditLine/CreditLine.sol
KalpitaMandal: Tried storing a local variable but was facing stack too deep error.

prateekreddy: We might have to update the code such that we can avoid stack too deep, by restricting scope of certain variables

KalpitaMandal: PR #275 invalidates this change. 
The above PR removes the return values sent by `transferFrom` function so the change made in this PR becomes invalid.
We might have to go back to the original version for this issue.
Need discussion.

prateekreddy (approved):

PR #275, opened 2022-01-11
CS lime-216, Unchecked Return Values

KalpitaMandal:
# Description

Multiple functions return a value which is not checked or used in the calling function. e.g.: 

• withdrawFrom in _withdrawBorrowAmount. 

• transfer and withdraw in _transferCollateral. 

• deposit in _repay. 

• transferFrom in _depositCollateralFromSavingsAccount and _repayFromSavingsAccount.

# Integrations Checklist

# Changelog

- Removing unwanted returns from the function `transferFrom` from the `SavingsAccount.sol`

# Event Signature Changes

# Function Signature Changes

# Features

# Events


prateekreddy (changes requested):
This issue need to be fixed as this function is also used for directly depositing to savings account, in which case a return value might be necessary

KalpitaMandal:
> This issue need to be fixed as this function is also used for directly depositing to savings account, in which case a return value might be necessary

The return value in `directDeposit` does not depend on `transferFrom`, rather the amount specified in the arguments is returned.

prateekreddy:
Current implementation of return values is inconsistent. `transfer` and `transferFrom` has different return values.

prateekreddy (approved):

PR #276, opened 2022-01-11
CS lime-217, Unnecessary Storage Read

KalpitaMandal:
# Description
In SavingsAccount.withdrawAll the _amount is set to balanceInShares[msg.sender][_token][_strategyList[i]]. If the condition _strategyList[i] != address(0) is true, it is immediately reassigned to another value. This first storage read could be avoided. As the read happens in a loop it might make sense to make the code more efficient.

# Integrations Checklist

# Changelog

- Removing redundant code in `withdrawAll` function in `SavingsAccount.sol`

# Event Signature Changes

# Function Signature Changes

# Features

# Events


contracts/SavingsAccount/SavingsAccount.sol
```diff
@@ -287,10 +287,7 @@ contract SavingsAccount is ISavingsAccount, Initializable, OwnableUpgradeable, R
 
         for (uint256 i = 0; i < _strategyList.length; i++) {
             if (balanceInShares[msg.sender][_token][_strategyList[i]] != 0) {
-                uint256 _amount = balanceInShares[msg.sender][_token][_strategyList[i]];
-                if (_strategyList[i] != address(0)) {
```
prateekreddy, line 291: I think what they meant here is inside the  if condition balanceInShares thing can be replaced by _amount

prateekreddy (changes requested):

prateekreddy (approved):

PR #282, opened 2022-01-11
CS lime-158, Wrong Calculation of Collateral Ratio in CreditLine

KalpitaMandal:
# Description
The borrow function in CreditLine computes the collateral ratio before an _amount is borrowed as follows: 
```
uint256 _collateralRatioIfAmountIsWithdrawn = _ratioOfPrices.mul(_totalCollateralToken) .div((_currentDebt.add(_amount)).mul(10**_decimals)); 
```
A simplified version of the above formula with corresponding decimals in square brackets is: (_ratioOfPrices[30] * _totalCollateralToken[x] * 10^30)/(_currentDebt + _amou nt)[y], where x is the decimals of the collateral token and y is the decimals of the borrow token. The result is in other decimals than 30.

# Integrations Checklist

# Changelog

- Adding correct calculation for collateral ratio.

# Event Signature Changes

# Function Signature Changes

# Features

# Events


contracts/CreditLine/CreditLine.sol
```diff
@@ -690,7 +690,7 @@ contract CreditLine is ReentrancyGuard, OwnableUpgradeable {
 
         uint256 _totalCollateralToken = calculateTotalCollateralTokens(_id);
 
-        uint256 _collateralRatioIfAmountIsWithdrawn = _ratioOfPrices.mul(_totalCollateralToken).div(
+        uint256 _collateralRatioIfAmountIsWithdrawn = _ratioOfPrices.mul(_totalCollateralToken).mul(10**30).div(
```
prateekreddy, line 693: Need to handle possible  overflow

prateekreddy (changes requested):

contracts/CreditLine/CreditLine.sol
KalpitaMandal: The variable was removed and no overflow detected for the new function flow

prateekreddy:
The borrow function uses a different way to calculate borrowable amount and hence the above logic is not used anymore.

PR #285, opened 2022-01-11
CS lime-241, Reentrancy Enabled in the Token Transfer

KalpitaMandal:
# Description
The function _beforeTokenTransfer in the Pool contract calls twice the function _withdrawRepayment for addresses _from and _to of the transfer. This enables the reentrancy as the function triggers .call to the above addresses. The transfer of the pool tokens has a critical vulnerability, refer to Incorrect Repayment Distribution, and our analysis of the impact of the reentrancy did not uncover other issues beyond this one. Furthermore, ERC777 tokens provide another possibility for reentrancy and the transfer of such tokens should be handled with special care.

# Integrations Checklist

# Changelog

- Adding Re-entrancy guard to `_withdrawRepayment` funcion in `Pool.sol`.

# Event Signature Changes

# Function Signature Changes
- `function _withdrawRepayment(address _lender) internal *nonReentrant*`

# Features

# Events


contracts/Pool/Pool.sol
```diff
@@ -944,7 +944,7 @@ contract Pool is Initializable, ERC20PausableUpgradeable, IPool, ReentrancyGuard
      * @notice internal function used to withdraw borrow asset from the pool by _lender
      * @param _lender address of the _lender
      */
-    function _withdrawRepayment(address _lender) internal {
+    function _withdrawRepayment(address _lender) internal nonReentrant {
```
ritik99, line 947: can you shift the nonReentrant guard to the `_beforeTokenTransfer` function? Because `_withdrawRepayment` is already called internally in functions with the nonReentrant guard which will cause problems

PR #289, opened 2022-01-14
CS lime-203, Not Accounted Ether Transfers

KalpitaMandal:
# Description
Multiple times the code checks whether the asset address is address zero. If so, it checks if the ether amount equals the provided amount argument `msg.value == amount`. In the case of token operations, it is not checked if Ether was sent in the same transaction. Ether might be sent accidentally in the transaction as the functions are payable. That Ether would not be accounted for.

# Integrations Checklist

# Changelog

- Assigning `msg.value` to zero in the following contracts, where ETH transfers are not required:
  - `CreditLine.sol`
  - `Pool.sol`
  - `Repayments.sol`
  - `SavingsAccount.sol`

# Event Signature Changes

# Function Signature Changes

# Features

# Events


contracts/CreditLine/CreditLine.sol
```diff
@@ -615,6 +615,9 @@ contract CreditLine is ReentrancyGuard, OwnableUpgradeable {
         bool _fromSavingsAccount
     ) external payable nonReentrant ifCreditLineExists(_id) {
         require(creditLineVariables[_id].status == CreditLineStatus.ACTIVE, 'CreditLine not active');
+        if(Address(creditLineConstants[_id].collateralAsset) != Address(0)) {
+            require(msg.value == 0, 'DepositCollateral: ETH is not required for this operation');
```
prateekreddy, line 619: This can be move to _depositCollateral function for gas optimization

prateekreddy:
Missed  this check in `_repay` function from creditLines

contracts/Pool/Pool.sol
```diff
@@ -172,6 +172,9 @@ contract Pool is Initializable, ERC20PausableUpgradeable, IPool, ReentrancyGuard
     function depositCollateral(uint256 _amount, bool _transferFromSavingsAccount) external payable override {
         require(_amount != 0, 'DC1');
         require(balanceOf(msg.sender) == 0, 'DC2');
+        if(Address(poolConstants.collateralAsset) != Address(0) || _transferFromSavingsAccount) {
```
prateekreddy, line 175: These checks can be move to the savingsAccountUtil functions which will make it more efficient

contracts/SavingsAccount/SavingsAccount.sol
```diff
@@ -112,6 +112,9 @@ contract SavingsAccount is ISavingsAccount, Initializable, OwnableUpgradeable, R
         address _to
     ) external payable override nonReentrant returns (uint256) {
         require(_to != address(0), 'SavingsAccount::deposit receiver address should not be zero address');
+        if(Address(_token) != Address(0)) {
+            require(msg.value == 0, 'Deposit: ETH is not required for this operation');
```
prateekreddy, line 116: May  be adding this to depositToYield internal function is more efficient

prateekreddy (changes requested):

KalpitaMandal:
> Missed this check in `_repay` function from creditLines

The check is present in `repay` function which calls the `_repay` function, so the same was not iterated in the internal function. 
Let me know if it i required anyway.

contracts/CreditLine/CreditLine.sol
KalpitaMandal: The change is made but having the check in the internal function will revert only after another function is called. Putting it in the external function will ensure the internal function is never called if ETH is sent.

contracts/Pool/Pool.sol
KalpitaMandal: Added the change

contracts/SavingsAccount/SavingsAccount.sol
KalpitaMandal: Added the change.

prateekreddy (changes requested):
https://github.com/sublime-finance/sublime-v1/blob/CS-lime-203/contracts/SavingsAccount/SavingsAccountUtil.sol#L49 doesn't have `msg.value` check and https://github.com/sublime-finance/sublime-v1/blob/CS-lime-203/contracts/Pool/Pool.sol#L594 function doesn't need to be payable. Receive method can be removed from all contracts. RepayPrinciple in repayments.sol hasn't been covered.

prateekreddy (approved):

PR #299, opened 2022-01-31
Fixing re-entrancy guard issue

KalpitaMandal:
Fixing the PR: https://github.com/sublime-finance/sublime-v1/pull/285


contracts/Pool/Pool.sol
```diff
@@ -444,6 +444,14 @@ contract Pool is Initializable, ERC20PausableUpgradeable, IPool, ReentrancyGuard
         if (_from == address(0) || _to == address(0)) {
             return;
         }
+        _pendingInterestWithdrawal(_from, _to, _amount);
+    }
+
+    function _pendingInterestWithdrawal(
```
prateekreddy, line 450: Rename to something more appropriate

prateekreddy (changes requested):

PR #310, opened 2022-02-11
fixes for lime-171

saxenism:
## Description (in branch v1-alpha)

• The specification of the priceOracle refers to the pool. 

• calculateRepaymentWithdrawable is declared as public, but the description refers to it as internal.


ritik99:
@prateekreddy we should keep calculateRepaymentWithdrawable as public and change the description right?

PR #314, opened 2022-02-18
Track Retired strategies

prateekreddy:
Strategy registry tracks the strategies which are removed to ensure that they can still be withdrawn from even after retiring.

### DIFF
--- a/contracts/interfaces/IStrategyRegistry.sol
+++ b/contracts/interfaces/IStrategyRegistry.sol
@@ -22,7 +22,7 @@ interface IStrategyRegistry {
 
     function registry(address strategy) external view returns (bool isValidStrategy);
 
-    function wasStrategy(address strategy) external view returns(bool wasValidStrategy);
+    function isValidStrategy(address strategy) external view returns(bool wasValidStrategy);
 
     function getStrategies() external view returns (address[] memory strategies);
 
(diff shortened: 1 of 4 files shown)


contracts/interfaces/IStrategyRegistry.sol
```diff
@@ -20,21 +20,23 @@ interface IStrategyRegistry {
      */
     event MaxStrategiesUpdated(uint256 maxStrategies);
 
-    function registry(address _strategy) external view returns (bool isValidStrategy);
+    function registry(address strategy) external view returns (bool isValidStrategy);
+
+    function isValidStrategy(address strategy) external view returns(bool wasValidStrategy);
```
ritik99, line 25: rename return var to isValidStrategy

contracts/interfaces/IStrategyRegistry.sol
```diff
@@ -20,21 +20,23 @@ interface IStrategyRegistry {
      */
     event MaxStrategiesUpdated(uint256 maxStrategies);
 
-    function registry(address _strategy) external view returns (bool isValidStrategy);
+    function registry(address strategy) external view returns (bool isValidStrategy);
```
ritik99, line 23: rename return var to isActiveStrategy

contracts/yield/StrategyRegistry.sol
```diff
@@ -83,6 +88,7 @@ contract StrategyRegistry is Initializable, OwnableUpgradeable, IStrategyRegistr
         strategies[_strategyIndex] = strategies[strategies.length.sub(1, 'StrategyRegistry::removeStrategy - No strategies exist')];
```
ritik99, line 88: for loop for removing strats so that relative order can be maintained

PR #315, opened 2022-02-18
Infinite approvals for creditLine from Lender

prateekreddy:
With the addition of pooledCreditLines, it is necessary for multiple addresses to be able to replenish their allowance when credit lines are repaid, hence support added for multiple credit lines.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,9 @@ test.spec.ts
 node_modules
 .env
 gasReport.md
+
+out
+out/*
+
+lib
+lib/*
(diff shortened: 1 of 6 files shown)


prateekreddy:
@ritik99 One of the concerns with this feature is any compromise to admin key would means all the tokens in savings account and thus all yields can be compromised.

Need to figure out a way to replenish credit line without giving an option to reset it.

One way to do this is infinite approval, so if infinite approvals can be provided to creditLine contracts by lender then the allowance to credit line can always stay during borrow and repay.

prateekreddy:
It has been decided post discussions that we will use infinite approvals and remove the `increaseAllowanceForCreditLine` function

ritik99 (approved):

PR #330, opened 2022-03-01
add blacklist twitterVerifier

iamzubin:
Add a function to blacklist digest


contracts/Verification/twitterVerifier.sol
```diff
@@ -168,4 +168,13 @@ contract TwitterVerifier is Initializable, IVerifier, OwnableUpgradeable, EIP712
...
         emit SignerUpdated(signerAddress);
     }
+
+    /**
+     * @notice blacklist a digest
+     * @dev only owner can update
+     * @param _hash digest hash to be blacklisted
+     */
+    function blackListDigest(bytes32 _hash) external onlyOwner {
+        hashAddressMap[_hash] = address();
```
prateekreddy, line 178: this is not right, may be use some random address like fffffff wala address

prateekreddy (changes requested):

contracts/Verification/twitterVerifier.sol
iamzubin: address(0) won't work?

prateekreddy: address 0 is default right ?

prateekreddy (approved):

PR #337, opened 2022-03-09
Changes in pooledCreditLine

saxenism:
1. Removed the `require` statement in the function `close` and replaced it with a pre-existing modifier
2. Added the `cancel` function in `PooledCreditLine.sol`
3. Added an event, emitted when a pooledCreditLine is cancelled
4. Removed an existing event which was supposed to be called when pooledCreditLine was `reset` (which never happens in the code base)
5. Added a function called `requestCancelled` in `LenderPool.sol` so that `creditLines` mapping for that particular `id` can be deleted too


contracts/PooledCreditLine/PooledCreditLine.sol
```diff
@@ -871,15 +870,27 @@ contract PooledCreditLine is ReentrancyGuard, OwnableUpgradeable, IPooledCreditL
      * @dev used to close credit line by borrower
      * @param _id identifier for the credit line
      */
-    function close(uint256 _id) external {
-        require(msg.sender == creditLineConstants[_id].borrower, 'C1');
-        require(creditLineVariables[_id].status == CreditLineStatus.ACTIVE, 'C2');
-        require(creditLineVariables[_id].principal == 0, 'C3');
+    function close(uint256 _id) external onlyCreditLineBorrower(_id) {
+        require(creditLineVariables[_id].status == CreditLineStatus.ACTIVE, '!ACTIVE');
```
prateekreddy, line 874: Error  messages  shouldn't  be changed

contracts/PooledCreditLine/PooledCreditLine.sol
```diff
@@ -871,15 +870,27 @@ contract PooledCreditLine is ReentrancyGuard, OwnableUpgradeable, IPooledCreditL
...
         emit CreditLineClosed(_id);
     }
 
+    /**
+     * @notice used to cancel pooledCreditLine request
+     * @dev only callable by the borrower
+     * @param _id identifier for the pooledCreditLine
+     */
+    function cancel(uint256 _id) external onlyCreditLineBorrower(_id) {
+        require(creditLineVariables[_id].status == CreditLineStatus.REQUESTED);
```
prateekreddy, line 887: needs message for require

contracts/PooledCreditLine/LenderPool.sol
```diff
@@ -108,6 +108,10 @@ contract LenderPool is ERC1155Upgradeable, ReentrancyGuard {
         }
     }
 
+    function requestCancelled(uint256 _id) external onlyPooledCreditLine {
+        delete creditLines[_id];
```
prateekreddy, line 112: If some lender has deposited tokens for lending, how will they withdraw ?

prateekreddy (changes requested):
Please look at above comments

saxenism:
> Please look at above comments

Resolved. Please review again.

contracts/PooledCreditLine/PooledCreditLine.sol
saxenism: added.

contracts/PooledCreditLine/PooledCreditLine.sol
saxenism: i was forking an older version of v1-alpha. pulled the latest version and fixed this. please review again.

contracts/PooledCreditLine/LenderPool.sol
saxenism: Updated.

Ok, so the approach I've followed now is:
1. `cancelRequest` is called by the borrower
2. It deletes creditLineConstants and creditLineVariables inside of PooledCreditLine.sol
3. This effectively makes the creditLineStatus as 0 (NOT_CREATED)
4. The `cancelRequest` function triggers `requestCancelled` function inside of `LenderPool`
5. The `requestCancelled` function set the `creditLines[_id].startTime as 0 (and does nothing else)
6. Now, I've added another conditional in the `withdrawLiquidity` function, so that the lenders can withdraw their liquidity in the case of credit line cancellation.

Essentially, I've opted for the lenders to withdraw their liquidity themselves instead of distributing the liquidity to all eligible lenders.

prateekreddy (approved):

PR #380, opened 2022-04-07
Fix withdrawable collateral in case of liquidations

prateekreddy:


KalpitaMandal:
In case of `EXPIRED` status, if a borrower has not borrowed anything from the PooledCreditline and has deposited collateral. He should be able to get his totalCollateral tokens back, right? But in the current scenario, the user will not be able to get back his collateral if he calls withdrawCollateral function. Is that the desired scenario or does the borrower have to go and do all the pool operations (borrow, repay, etc) to get back his collateral?

prateekreddy:
> In case of `EXPIRED` status, if a borrower has not borrowed anything from the PooledCreditline and has deposited collateral. He should be able to get his totalCollateral tokens back, right? But in the current scenario, the user will not be able to get back his collateral if he calls withdrawCollateral function. Is that the desired scenario or does the borrower have to go and do all the pool operations (borrow, repay, etc) to get back his collateral?

True.... I have allowed closing of pool in expired state as well as long as principal is 0. That should fix this.

ritik99 (approved):
Merge conflicts need to be resolved

KalpitaMandal (approved):
lgtm

ritik99:
lgtm

PR #381, opened 2022-04-07
use shares denominated amount in terminate _principalWithdrawable

prateekreddy:


KushGoyal:
Hey please check the comments I have added here
https://github.com/sublime-finance/sublime-v1/blob/kush/c4march21-copy/contracts/PooledCreditLine/LenderPool.sol

saxenism:
> Hey please check the comments I have added here https://github.com/sublime-finance/sublime-v1/blob/kush/c4march21-copy/contracts/PooledCreditLine/LenderPool.sol#L357

- https://github.com/sublime-finance/sublime-v1/blob/kush/c4march21-copy/contracts/PooledCreditLine/LenderPool.sol#L360
  -- It is not necessary that `_borrowLimit` is the *borrow amount asked by the borrower in the PCL*. For example, if just the `_minBorrowAmount` was met and the `start` function was called.
- https://github.com/sublime-finance/sublime-v1/blob/kush/c4march21-copy/contracts/PooledCreditLine/LenderPool.sol#L364
  -- Typo: `_borrowLimit` and not `_borrowedLimit`. Also instead of `principal borrowed` it would be better using `amount borrowed`, since you anyway can't borrow anything apart from the principal
- https://github.com/sublime-finance/sublime-v1/blob/kush/c4march21-copy/contracts/PooledCreditLine/LenderPool.sol#L365

KushGoyal:
@saxenism 

I have made some changes in the comments. But I have not understood the issue with line 373-375.

as stated 

> // the amount of _borrowAsset deposited by lenders against the PCL is equal to totalSupply[_id]
   // when the PCL is active totalSupply[_id] == _borrowedLimit
   // but _borrowLimit is some cases can be different than totalSupply[_id]
   // for eg: before the PCL has started the borrowLimit asked by borrower is 1000USDC
   // but the lenders have only deposited 900USDC till now so totalSupply[_id] is 900USDC

borrowLimit is set twice, once in the createRequest method and then in the _accept method. The above statements seem true to me.

saxenism (approved):

ritik99 (approved):

PR #382, opened 2022-04-07
fix for c4march-20 lime-290

KushGoyal:
https://linear.app/sublime-finance/issue/LIME-290/20-expected-withdrawn-interest-shares-might-not-match-actual-amount


prateekreddy:
@KushGoyal We can probably remove _interestToWithdraw as one of the return values as it is not being used in this function.

contracts/PooledCreditLine/LenderPool.sol
```diff
@@ -393,17 +393,18 @@ contract LenderPool is ERC1155Upgradeable, ReentrancyGuard, IPooledCreditLineEnu
         address _strategy = pooledCLConstants[_id].borrowAssetStrategy;
         address _borrowAsset = pooledCLConstants[_id].borrowAsset;
 
-        (uint256 _interestToWithdraw, uint256 _interestSharesToWithdraw) = _calculateInterestToWithdraw(
+        (, uint256 _interestSharesToWithdraw) = _calculateInterestToWithdraw(
```
prateekreddy, line 396: We can remove the return value from the definition of _calculateInterestToWithdraw ?

KushGoyal: this is being used in another place with both the values
https://github.com/sublime-finance/sublime-v1/blob/d75a260ab2b6c1ce46661ff5c9ad982427fea2dd/contracts/PooledCreditLine/LenderPool.sol#L543

prateekreddy: right, but we can seperate that out to save gas here.

PR #388, opened 2022-04-15
Lime-381 Misleading Variable Name _numTokens

saxenism:
The variable _numTokens in getUniswapLatestPrice stores how many wei``s of ``den token match 1 wei of num token.


contracts/PriceOracle.sol
```diff
@@ -120,8 +120,8 @@ contract PriceOracle is Initializable, OwnableUpgradeable, IPriceOracle {
         if (_pool == address(0)) return (0, 0);
 
         int24 _twapTick = OracleLibrary.consult(_pool, uniswapPriceAveragingPeriod);
-        uint256 _numTokens = OracleLibrary.getQuoteAtTick(_twapTick, uint128(SCALING_FACTOR), num, den);
-        return (_numTokens, SCALING_EXPONENT);
+        uint256 _numTokenWeiInTermsOfDenToken = OracleLibrary.getQuoteAtTick(_twapTick, uint128(SCALING_FACTOR), num, den);
```
prateekreddy, line 123: Rename it to price probably and change natspec to tell that price is in terms of wei for num and denom tokens

prateekreddy (changes requested):

contracts/PriceOracle.sol
saxenism: Implemented this change. 

Please review again.

PR #390, opened 2022-04-15
Lime-371: Description of _asset Address in CompoundYield

saxenism:
The descriptions of the parameter _asset in the functions unlockTokens and unlockShares could be more precise.


contracts/yield/CompoundYield.sol
```diff
@@ -176,7 +176,7 @@ contract CompoundYield is IYield, Initializable, OwnableUpgradeable, ReentrancyG
 
     /**
      * @notice Used to unlock tokens from available protocol
-     * @param _asset the address of share token
+     * @param _asset the address of share token (Example: For Compound, Share token of TokenA is cTokenA)
```
prateekreddy, line 179: can you recheck ? `_asset` is address of underlying token not the cToken. Also the description of asset in unlockShares doesn't seem right

prateekreddy (changes requested):

contracts/yield/CompoundYield.sol
saxenism: True.
I think the definition in both those functions got swapped. Have corrected it now.

Please review again.

prateekreddy (approved):

